PR #5, opened 2016-01-14
Moved the hosts method to the Api class

syncrou:


syncrou:
@bdunne - Please review

PR #8, opened 2016-02-02
Exposed the passing extra_vars to launch

syncrou:
@bdunne - (ping) Error handling for bad vars will be in the next PR.


lib/ansible_tower_client/job_template.rb
```diff
@@ -3,8 +3,15 @@ class JobTemplate
     extend CollectionMethods
     include InstanceMethods
 
-    def launch
-      Api.post("#{url}launch/")
+    attr_reader :response
+
+    def launch(vars = {})
+      launch_url = "#{url}launch/"
+      @response = Api.post(launch_url, vars).body
```
bdunne: @syncrou Sorry for the delay, I'm okay with merging this for now, but I feel like this should probably return a Job object.  What do you think?

syncrou: Yes - I like that idea.  I sort of forgot that we could actually build a real class around that response. I'll get that added.

lib/ansible_tower_client/job_template.rb
```diff
@@ -3,8 +3,15 @@ class JobTemplate
...
+    end
+
+    def extra_vars
```
bdunne: Can this be set up on initialize with an `attr_reader` instead?

syncrou:
@bdunne - Just pushed the updates for you to review.

bdunne:
Thanks! :+1:

PR #10, opened 2016-02-09
Add validation

syncrou:


syncrou:
@bdunne - Please review

lib/ansible_tower_client/valid_json.rb
```diff
@@ -0,0 +1,12 @@
+module AnsibleTowerClient
+  module ValidJSON
+    def validate(*args)
+      begin
+        JSON.parse(args.first)
```
bdunne: Why `*args` above and `args.first` here?  Isn't it always a single string being passed in?

syncrou:
@bdunne - Fixed the *args argument.

PR #12, opened 2016-02-11
Spec cleanup

bdunne:
Ensure collections are built with the correct type.

Fix lots of spec failures due to this change.
Add FactoryGirl and reduce duplication through factories and shared_examples

### DIFF
--- a/spec/group_spec.rb
+++ b/spec/group_spec.rb
@@ -5,28 +5,7 @@ describe AnsibleTowerClient::Group do
   let(:collection)     { build(:response_collection, :klass => described_class) }
   let(:instance)       { build(:response_instance,   :klass => described_class) }
 
-  context "Collection Methods" do
-    describe '.all' do
-      let(:get) { instance_double("Faraday::Result", :body => collection.to_json) }
-
-      it "returns a collection of objects" do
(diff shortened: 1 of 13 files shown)


bdunne:
@syncrou Please review.
cc @gmcculloug 
+ @jrafanie for the :scissors: :fire: :smile:

spec/group_spec.rb
```diff
@@ -1,50 +1,17 @@
...
-                   :url => '/api/v1/groups/1/', :name => 'test1'},
-                  {:id => 2, :type => 'groups',
-                   :url => '/api/v1/groups/2/', :name => 'test2'}]}.to_json
-  end
-
-  let(:one_result) do
-    {:id => 1, :url => '/api/v1/groups/1/', :name => 'test1'}.to_json
-  end
-
   let(:api_connection) { instance_double("Faraday::Connection", :get => get) }
```
syncrou: Do we need the get method as part of this instance_double? (on any of the specs that don't actually call ':get')

syncrou:
:+1:  - minus the ':get' question.

bdunne:
The `include_examples "Collection Methods"` still depends on that.

PR #14, opened 2016-02-12
Survey spec

syncrou:


syncrou:
@bdunne - I redid the validation process - I found I needed it to work dynamically with the survey_specs - so the survey spec update process will be dependent on this PR.

bdunne:
@syncrou Can you split this into a couple of PRs?  Gemspec changes, whitespace, leaving just the survey related work here?

syncrou:
@bdunne - Yes I'll get that out in the morning.

bdunne:
This validation logic seems complicated, maybe I don't understand all of the use cases.  I would expect `validate_json` to return a valid JSON string or raise.

The only caller that I am aware of right now is `extra = JSONValues.new(vars).extra_vars` in `JobTemplate.launch`

I think it would be a lot simpler to do something like:
```ruby
extra = Validation.validate_json(vars['extra_vars'])

module AnsibleTowerClient
  module Validation
    def self.validate_json(text)
      JSON.parse(text)
      text
    rescue JSON::ParserError => e
      raise InvalidJSON.new(self, e)
    end

    def self.validate_yaml(text)
      YAML.safe_parse(text)
      text
    rescue YAML::ParserError => e
      raise InvalidYAML.new(self, e)
    end
  end
end
```

PR #15, opened 2016-02-24
Extra vars changes in job_template

syncrou:
The extra_vars key needs to be passed to
Tower as a hash key with JSON values.

This resolves an issue where the passed in
extra_vars were not being applied to the
template being launched against.

### DIFF
--- a/lib/ansible_tower_client/job_template.rb
+++ b/lib/ansible_tower_client/job_template.rb
@@ -14,7 +14,8 @@ module AnsibleTowerClient
     def launch(vars = {})
       launch_url = "#{url}launch/"
       extra = JSONValues.new(vars).extra_vars
-      resp = Api.post(launch_url, extra).body
+      extra_vars = {"extra_vars": extra.to_json}
+      resp = Api.post(launch_url, extra_vars).body
       job = JSON.parse(resp)
       Job.find(job['job'])
     end


syncrou:
@bdunne - Please review

lib/ansible_tower_client/job_template.rb
```diff
@@ -14,7 +14,8 @@ def initialize(raw_body)
     def launch(vars = {})
       launch_url = "#{url}launch/"
       extra = JSONValues.new(vars).extra_vars
-      resp = Api.post(launch_url, extra).body
+      extra_vars = {"extra_vars": extra.to_json}
```
bdunne: @syncrou I think we need to be very careful here.  Our caller will probably send a JSON string in to the launch method and I think we'll be in trouble if we `.to_json` that again.

syncrou: @bdunne - Yeah - Then I'll have to fix the issue in JSONValues - I believe it is currently doing a JSON.parse on extra_vars and returning as a native ruby hash - which isn't correct. It probably should see if the parse doesn't blow up - and then return what was sent.

bdunne: @syncrou Yes, that sounds better

PR #16, opened 2016-03-01
Add survey spec

syncrou:
@bdunne - Please review

### DIFF
--- a/spec/factories/responses.rb
+++ b/spec/factories/responses.rb
@@ -15,7 +15,7 @@ FactoryGirl.define do
     type       { AnsibleTowerClient::FactoryHelper.underscore_string(klass.namespace.last) }
     name       { "#{type}-#{id}" }
     url        { "/api/v1/#{klass.endpoint}/#{id}/" }
-    related    { {"survey_spec" => ["blah blah"]} }
+    related    { {"survey_spec" => "example.com/api"} }
 
     trait(:description)  { sequence(:description) { |n| "description_#{n}" } }
     trait(:extra_vars)   { extra_vars "lots of options" }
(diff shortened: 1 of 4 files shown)


spec/support/factory_helper.rb
```diff
@@ -3,7 +3,8 @@
 module AnsibleTowerClient
   module FactoryHelper
     def self.stringify_attribute_keys(hash)
-      [:description, :extra_vars, :id, :instance_id, :inventory, :name, :results, :type, :url].each do |attribute|
+      [:description, :extra_vars, :id, :instance_id, :inventory, :name, :results,
```
bdunne: If the array needs to be wrapped, it should probably be in a constant at the top.

spec/factories/responses.rb
```diff
@@ -15,6 +15,7 @@
     type       { AnsibleTowerClient::FactoryHelper.underscore_string(klass.namespace.last) }
     name       { "#{type}-#{id}" }
     url        { "/api/v1/#{klass.endpoint}/#{id}/" }
+    related    { {"survey_spec" => ["blah blah"]} }
```
bdunne: Is it actually an array of strings?  Usually the "related" are links to other collections, right?

syncrou:
@bdunne - Made the requested changes.  Please review.

spec/job_template_spec.rb
```diff
@@ -31,4 +31,15 @@
...
   end
+
+  describe '#survey_spec' do
+    let(:get) { instance_double("Faraday::Result", :body => instance.to_json) }
+
+    it "returns a survey spec if it exists" do
+      AnsibleTowerClient::Api.instance_variable_set(:@instance, api_connection)
+      survey = described_class.new(instance).survey_spec
+      expect(survey).to be_a Hash
+      expect(survey['related']['survey_spec']).to eq 'example.com/api'
```
bdunne: I think this 'related' is wrong.  I think your get is returning a job_template rather than the survey response.

syncrou:
@bdunne - Pushed changes including if a survey_spec is nil

bdunne:
:+1:

PR #17, opened 2016-03-01
Add BaseModel class to help auto generate attribute accessors

bzwei:
The `BaseModel` class is borrowed from our armrest gem. It automatically converts all hash attributes to instance methods. It can preserve named list of hash attribute from being converted. So far I don't know which attributes need to remain as hash.

With this work module `InstanceMethods` is no longer needed.

Update: attributes that appear in the related field will be treated as ids. So job_template, inventory, cloud_credential, project etc will be automatically appended with _id.


gmcculloug:
@bdunne @syncrou Please review.

lib/ansible_tower_client/group.rb
```diff
@@ -1,14 +1,6 @@
 module AnsibleTowerClient
-  class Group
+  class Group < BaseModel
     extend CollectionMethods
-    include InstanceMethods
-
-    attr_reader :inventory_id
-
-    def initialize(raw_body)
-      @inventory_id = raw_body["inventory"]
```
syncrou: How do we implement a mapping back to \@inventory_id?

bzwei: You can define your own method for particular model, for example
```
def inventory_id
  __getobj__['inventory']
end
```
The auto generated `inventory` method still exists and returns the same thing.

syncrou:
:+1:  LGTM - Except in regards to the removed inventory_id attribute for Group, Host - I thought the app was expecting a return on that method. If not - I'm set.

bzwei:
@syncrou If `inventory_id` is already used in manageiq app, then we need to update manageiq also. Or we can add back `inventory_id` method.

bdunne:
@bzwei Should we extract [MiqHashStruct](https://github.com/ManageIQ/manageiq/blob/master/gems/pending/util/miq-hash_struct.rb) to more_core_extensions and use it as a superclass for BaseModel?  I think that would simplify the code in BaseModel.

bdunne:
@bzwei I'm still stuck on `AnsibleTowerClient::Host#inventory.class` returning `Integer` rather than `AnsibleTowerClient::Inventory`.

bzwei:
@bdunne @syncrou Update: attributes that appear in the `related` field will be treated as ids. So `job_template`, `inventory`, `cloud_credential`, `project` etc will be automatically appended with `_id`.

bzwei:
@bdunne @syncrou please review and merge.

syncrou:
@bzwei - :+1:

lib/ansible_tower_client/base_model.rb
```diff
@@ -1,9 +1,152 @@
...
+    #   person = Person.new(json_string)
+    #
+    #   # The JSON properties are now available as methods.
+    #   person.firstname        # => 'jeff'
+    #   person.address.zipcode  # => '01013'
+    #
+    #   # Or you can get back the original JSON if necessary.
+    #   person.to_json # => Returns original JSON
+    #
+    def initialize(api, json = {})
```
bdunne: Why is `json` optional?

lib/ansible_tower_client/base_model.rb
```diff
@@ -1,9 +1,152 @@
...
+      # '#' is the separator between levels. Remove attributes
+      # before the first separator.
+      @child_excl_list = self.class.send(:excl_list).map do |e|
+        e.index('#') ? e[e.index('#') + 1..-1] : ''
+      end
+
+      __setobj__(json.kind_of?(Hash) ? json.dup : JSON.parse(json))
+    end
+
+    def to_h
```
bdunne: Should this be an alias?

lib/ansible_tower_client/base_model.rb
```diff
@@ -1,9 +1,152 @@
...
+
+    def to_hash
+      __getobj__.dup
+    end
+
+    def to_json
+      __getobj__.to_json
+    end
+
+    def to_s
```
bdunne: Should this be an alias?

lib/ansible_tower_client/base_model.rb
```diff
@@ -1,9 +1,152 @@
...
+        child_excl_list = @child_excl_list
+        self.class.const_set(klass_name, Class.new(BaseModel) { attr_hash(*child_excl_list) })
+      end
+      self.class.const_get(klass_name).new(api, value)
+    end
+
+    def add_accessor_methods(method, key)
+      method.prepend('_') if methods.include?(method.to_sym)
+      instance_eval { define_singleton_method(method) { __getobj__[key] } }
+      instance_eval { define_singleton_method("#{method}=") { |val| __getobj__[key] = val } }
```
bdunne: If i'm reading this right, we will define `"#{method}=` for all attributes that will write to the objects underlying `@hashobj`, but not make any calls to the REST API.  I would expect `"#{method}=` to call the REST API to update an attribute then probably refresh the underlying `hashobj`, I don't think we should dynamically define all of those methods.

lib/ansible_tower_client/base_model.rb
```diff
@@ -1,9 +1,152 @@
...
+      return false unless other.kind_of?(BaseModel)
+      __getobj__.eql?(other.__getobj__)
+    end
+
+    # Support hash style accessors
+    def [](key)
+      __getobj__[key]
+    end
+
+    def []=(key, val)
```
bdunne: Same concern as https://github.com/ManageIQ/ansible_tower_client/pull/17/files#r55891091

spec/support/factory_helper.rb
```diff
@@ -2,8 +2,8 @@
 
 module AnsibleTowerClient
   module FactoryHelper
-    KEYS = [:description, :extra_vars, :id, :instance_id,
-            :inventory, :name, :results, :type, :url, :related].freeze
+    KEYS = [:related, :description, :extra_vars, :id, :instance_id,
```
bdunne: Is there a reason this can't be in alphabetical order?

lib/ansible_tower_client/base_model.rb
```diff
@@ -1,9 +1,152 @@
...
+      self.class.const_get(klass_name).new(api, value)
+    end
+
+    def add_accessor_methods(method, key)
+      method.prepend('_') if methods.include?(method.to_sym)
+      instance_eval { define_singleton_method(method) { __getobj__[key] } }
+      instance_eval { define_singleton_method("#{method}=") { |val| __getobj__[key] = val } }
+    end
+
+    def snake_case(name)
```
bdunne: With ActiveSupport, you should be able to remove this in favor of `"".underscore`.

bdunne:
@bzwei I think this PR is getting much closer.  I still think we should split `BaseModel` into:
- `BaseModel` with an initialize method that filters what attribute methods should be defined and
- its underlying `ReadOnlyHashStruct`

Does anyone else have an opinion? @Fryguy @gmcculloug @syncrou

bdunne:
Also, If BaseModel is split, the ReadOnlyHashStruct can do something very similar to MiqHashStruct using `method_missing` and `respond_to_missing?` to hit`:[]`.  I think that will simplify the code a lot and remove the need for the `instance_eval` and `define_method` lines.

lib/ansible_tower_client/base_model.rb
bzwei: The model class when used in armrest gem does not contain the service class/provider connection. It services purely as an attributes container. Therefore it makes sense to support both read and write.

In this gem, if you expect any changes in the model to be in sync with its corresponding provider object, then we should not auto_generate the setter.

spec/support/factory_helper.rb
bzwei: `:related` has to be the first because its entries decide what attributes needs to be appended with `_id`

bzwei:
@bdunne the nice thing with this implementation is that it is more like a real class. You can call #methods(false) to get all new methods it added based on the hash.

spec/support/factory_helper.rb
bdunne: I wouldn't rely on it being first in the response.

PR #18, opened 2016-03-03
Return survey_spec as a string

syncrou:
We currently return extra_vars as a raw string.
Survey_spec was returned as a JSON.parsed Hash.
The app is expecting to save a raw string - identical to extra_vars
This simply makes the two attributes more alike.


syncrou:
@bdunne - Please review

PR #20, opened 2016-03-09
Allow validation on multiple fields with different text structures

syncrou:
All keys passed into the Data.validate! method
can now be correctly parsed for valid Yaml or JSON.

Once parsed they are reoriented to match the
tower data structure so the API can process the requested
changes.


syncrou:
@bdunne @bzwei - Please review

PR #21, opened 2016-03-11
Allow validation on multiple fields with different text structures

syncrou:
All keys passed into the Data.validate! method
can now be correctly parsed for valid Yaml or JSON.

Once parsed they are reoriented to match the
tower data structure so the API can process the requested
changes.


syncrou:
@bdunne - Reopened after a rebase and a few small refactors in the Data class. -- cc @bzwei, @gmcculloug

bdunne:
@syncrou Do we need to pass in the whole hash rather than just the strings we need to validate?

syncrou:
@bdunne - My thought was to pass in the whole hash so we could eliminate logic in separating out specific keys.  

The other thought was that if we validate everything passed in (on the gem side) - the API will only use what it needs.  Not sure if that is a common pattern to follow - but would allow the API to adjust while not having to manage it all in our gem. 

I also wanted the Data class to 'do the right thing' and return back to the caller only what would be needed.

PR #22, opened 2016-03-16
Added with_temporary_changes to handle updating limit

syncrou:
1. Pass the post url with extra_vars into a block
2. Make a separate patch call to update the limit
3. Yield the original launch restful call
4. Ensure the limit is set back to what it originally was.

### DIFF
--- a/lib/ansible_tower_client/job_template.rb
+++ b/lib/ansible_tower_client/job_template.rb
@@ -1,10 +1,14 @@
 module AnsibleTowerClient
   class JobTemplate < BaseModel
-    def launch(vars = {})
+    def launch(options = {})
       launch_url = "#{url}launch/"
-      extra = JSONValues.new(vars).extra_vars
-      resp = api.post(launch_url, extra).body
-      job = JSON.parse(resp)
+      options = options.dup
+      new_limit = options.delete('limit')
+      response = with_temporary_changes(new_limit) do
+        api.post(launch_url, options).body
+      end
+
+      job = JSON.parse(response)
       api.jobs.find(job['job'])
     end
 
@@ -16,6 +20,27 @@ module AnsibleTowerClient
 
     def self.endpoint
       "job_templates".freeze
+    end
+
+    private
+
+    def with_temporary_changes(in_limit)
+      old_limit = limit
+      new_limit = in_limit
+      patch("{ \"limit\": \"#{new_limit}\" }")
+      begin
+        yield
+      ensure
+        patch("{ \"limit\": \"#{old_limit}\" }")
+      end
+    end
+
+    def patch(body)
+      api.patch do |req|
+        req.url(url)
+        req.headers['Content-Type'] = 'application/json'
+        req.body = body
+      end
     end
   end
 end
--- a/spec/factories/responses.rb
+++ b/spec/factories/responses.rb
@@ -16,6 +16,7 @@ FactoryGirl.define do
     name       { "#{type}-#{id}" }
     url        { "/api/v1/#{klass.endpoint}/#{id}/" }
     related    { {"survey_spec" => "example.com/api", 'inventory' => 'inventory link'} }
+    limit      { "default" }
 
     trait(:description)             { sequence(:description) { |n| "description_#{n}" } }
     trait(:extra_vars)              { extra_vars "lots of options" }
--- a/spec/job_template_spec.rb
+++ b/spec/job_template_spec.rb
@@ -22,13 +22,20 @@ describe AnsibleTowerClient::JobTemplate do
 
   describe '#launch' do
     let(:json) { {'extra_vars' => "{\"instance_ids\":[\"i-999c\"],\"state\":\"absent\",\"subnet_id\":\"subnet-887\"}"} }
+    let(:limit) { 'test' }
     let(:post_result_body) { {:job => 1} }
 
     it "runs an existing job template" do
       expect_any_instance_of(AnsibleTowerClient::Collection).to receive(:find).with(post_result_body[:job])
       expect(api).to receive(:post).and_return(instance_double("Faraday::Response", :body => post_result_body.to_json))
+      expect(api).to receive(:patch).twice.and_return(instance_double("Faraday::Response", :body => post_result_body.to_json))
 
       described_class.new(api, raw_instance).launch(json)
+    end
+
+    it "handles limit when passed in" do
+      expect_any_instance_of(AnsibleTowerClient::JobTemplate).to receive(:patch).twice
+      described_class.new(api, raw_instance).send(:with_temporary_changes, limit) { '' }
     end
   end
 


syncrou:
@bdunne, @gmcculloug - Please review - cc @bzwei @mkanoor

syncrou:

lib/ansible_tower_client/error.rb
```diff
@@ -32,8 +32,17 @@ def initialize(klass)
       @message = "Error on #{klass.class.name}: #{klass.inspect}"
       super(message)
     end
+
+    def self.parse(text)
+      txt = []
+      JSON.parse(text).each do |name, error|
```
bzwei:
```
JSON.parse(text).collect do ...
end.join("\n")
```

lib/ansible_tower_client/job_template.rb
```diff
@@ -10,12 +10,19 @@ def initialize(api, raw_body)
...
+      resp = with_temporary_changes(new_limit) do
+        api.post(launch_url, options).body
+      end
+
+      begin
+        job = JSON.parse(resp)
+        api.jobs.find(job['job'])
+      rescue ResourceNotFound
+        Error.parse(resp)
+      end
```
bzwei: What is the return on success and failure? Is it either a `Job` or an error string? An error should be thrown in case of failure.

lib/ansible_tower_client/job_template.rb
```diff
@@ -10,12 +10,19 @@ def initialize(api, raw_body)
       super
     end
 
-    def launch(vars = {})
+    def launch(options = {})
```
bzwei: I would prefer `launch(limit = nil, vars = {})`
If to use options, it should have format `:limit => 'some condition', :extra_vars => {the extra vars}`

lib/ansible_tower_client/job_template.rb
```diff
@@ -27,5 +34,28 @@ def survey_spec
...
+      rescue => e
+        raise InvalidPost.new(e)
+      ensure
+        patch("{ \"limit\": \"#{old_limit}\" }")
+      end
+    end
+
+    def patch(body)
+      api.patch do |req|
+        req.url url
```
bzwei: If we all follow the same style by manageiq, we should write as `req.url(url)`. I am surprised at how to setup the `url` vs `body` as the syntax is different.

ansible_tower_client.gemspec
```diff
@@ -35,4 +35,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "factory_girl"
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec"
+  spec.add_development_dependency "byebug"
```
bdunne: Please do this in a separate PR.

lib/ansible_tower_client/job_template.rb
bdunne: @bzwei I suggested `options` due to the number of optional args that could be passed in here.  In the future we'll also want to pass in survey data.

lib/ansible_tower_client/job_template.rb
```diff
@@ -10,12 +10,19 @@ def initialize(api, raw_body)
...
     end
 
-    def launch(vars = {})
+    def launch(options = {})
       launch_url = "#{url}launch/"
-      extra = JSONValues.new(vars).extra_vars
-      resp = api.post(launch_url, extra).body
-      job = JSON.parse(resp)
-      api.jobs.find(job['job'])
+      new_limit = options.delete('limit')
```
bdunne: @syncrou we may want to `options.dup` to prevent modifying the real options

lib/ansible_tower_client/job_template.rb
```diff
@@ -10,12 +10,19 @@ def initialize(api, raw_body)
...
+      resp = with_temporary_changes(new_limit) do
```
bdunne: No need to shorten `response`

lib/ansible_tower_client/job_template.rb
```diff
@@ -10,12 +10,19 @@ def initialize(api, raw_body)
...
+        api.post(launch_url, options).body
+      end
+
+      begin
+        job = JSON.parse(resp)
+        api.jobs.find(job['job'])
+      rescue ResourceNotFound
```
bdunne: Did we not hit this error in the past?

bdunne: This type of exception handling should probably be on the collection, not here.

lib/ansible_tower_client/job_template.rb
bdunne: Agreed, object or raise

lib/ansible_tower_client/job_template.rb
```diff
@@ -27,5 +34,28 @@ def survey_spec
...
+
+    private
+
+    def with_temporary_changes(limit)
+      old_limit = @raw_body['limit']
+      new_limit = limit
+      patch("{ \"limit\": \"#{new_limit}\" }")
+      begin
+        yield
+      rescue => e
```
bdunne: Don't rescue all exceptions.  I expect that any errors would be from the `api.post` [here](https://github.com/ManageIQ/ansible_tower_client/pull/22/files#diff-c799bcb40bc98d6ec3bcd2e4b41361e6R17) and only those should be rescued.

lib/ansible_tower_client/job_template.rb
syncrou: @bdunne - We did hit this in the past - I was trying to handle it better with the `Error.parse` method. In the past Faraday would raise ResourceNotFound - but the actual text of the that error - e.g. the API response was lost.  If I'm reading what you and @bzwei want - we still need to raise an error - but pass the API response as text shown in the error class message?

lib/ansible_tower_client/job_template.rb
syncrou: I was going to ask about that - I didn't feel great about leaving `rescue` open ended either but wasn't sure on the standard operating procedure.

syncrou:
@bdunne - @bzwei - I applied the requested changes - Please review.

lib/ansible_tower_client/error.rb
```diff
@@ -28,12 +28,20 @@ def general_msg
 
   class Error < StandardError
     attr_reader :message
-    def initialize(klass)
-      @message = "Error on #{klass.class.name}: #{klass.inspect}"
-      super(message)
+    def initialize(klass, message = nil)
+      @message = message.nil? ? "Error on #{klass.class.name}: #{klass.inspect}" : message
```
bdunne: Maybe `@message = message || "Error on #{klass.class.name}: #{klass.inspect}"`

lib/ansible_tower_client/job_template.rb
```diff
@@ -10,11 +10,15 @@ def initialize(api, raw_body)
       super
     end
 
-    def launch(vars = {})
+    def launch(options = {})
       launch_url = "#{url}launch/"
-      extra = JSONValues.new(vars).extra_vars
-      resp = api.post(launch_url, extra).body
-      job = JSON.parse(resp)
+      new_limit = options.dup.delete('limit')
```
bdunne: I think this needs to be:
```
options = options.dup
new_limit = options.delete('limit')
```

lib/ansible_tower_client/job_template.rb
```diff
@@ -10,11 +10,15 @@ def initialize(api, raw_body)
...
+      response = with_temporary_changes(new_limit) do
+        api.post(launch_url, options).body
+      end
+
+      job = JSON.parse(response)
+      raise InvalidOptions.new(self, Error.parse(response)) if job['job'].nil?
```
bdunne: I still don't know why we need this since we are handling this error [here](https://github.com/ManageIQ/ansible_tower_client/blob/master/lib/ansible_tower_client/collection.rb#L27)

syncrou: The thought was so we could see errors like this `AnsibleTowerClient::InvalidOptions: Error on variables_needed_to_start: ["'instance_ids' value missing", "'crap_id' value missing"]`

lib/ansible_tower_client/job_template.rb
bdunne: In this implementation, we still pass limit in the post options on line 17.  That was supposed to be removed from the options since the block already takes care of setting the limit for us.

```ruby
irb(main):001:0> raw_options = {:limit => "abc", :extra_vars => "xyz"}
=> {:limit=>"abc", :extra_vars=>"xyz"}
irb(main):002:0> options = raw_options
=> {:limit=>"abc", :extra_vars=>"xyz"}
irb(main):003:0> raw_options.object_id
=> 47375013423080
irb(main):004:0> options.object_id
=> 47375013423080
irb(main):005:0> limit = options.dup.delete(:limit)
=> "abc"
irb(main):006:0> options
=> {:limit=>"abc", :extra_vars=>"xyz"}
```

The `dup` needs to be on its own line because:
```ruby
irb(main):001:0> raw_options = {:limit => "abc", :extra_vars => "xyz"}
=> {:limit=>"abc", :extra_vars=>"xyz"}
irb(main):002:0> options = raw_options.dup
=> {:limit=>"abc", :extra_vars=>"xyz"}
irb(main):003:0> raw_options.object_id
=> 47002009217800
irb(main):004:0> options.object_id
=> 47002008338660
irb(main):005:0> limit = options.delete(:limit)
=> "abc"
irb(main):006:0> limit
=> "abc"
irb(main):007:0> options
=> {:extra_vars=>"xyz"}
irb(main):008:0> raw_options
=> {:limit=>"abc", :extra_vars=>"xyz"}
```

lib/ansible_tower_client/job_template.rb
bdunne: Thanks, the discussion cleared up a lot.  You're correct, this is a different problem from the resource not being found.
I was expecting Faraday to raise an error when it receives a HTTP 400 response from the `api.post` above.  I think we should probably handle this error generically in the Api class.  Let's do that in a followup PR.

syncrou:
@bdunne - Pushed those changes

bdunne:
:+1:

syncrou:
@bdunne - rebased

PR #24, opened 2016-03-29
Handle limit when it is passed in as a symbol

syncrou:


syncrou:
@miq-bot add_label bug

syncrou:
@bdunne - please review.

lib/ansible_tower_client/job_template.rb
```diff
@@ -3,7 +3,7 @@ class JobTemplate < BaseModel
     def launch(options = {})
       launch_url = "#{url}launch/"
       options = options.dup
-      new_limit = options.delete('limit')
+      new_limit = options.delete('limit') { options.delete(:limit) }
```
gmcculloug: @syncrou For something this simple I think this reads better: 
`options.delete('limit') || options.delete(:limit)`  (If you are expecting a symbol to be passed from our code it would be good to swap the order as well.)

If you needed to use the block I like the idea of passing the key into and calling `to_sym`.
`options.delete('limit') { |key| options.delete(key.to_sym) }`

gmcculloug:
Looks good, just needs a test.

gmcculloug:
@bdunne please review.

PR #25, opened 2016-04-01
Raise error middleware

syncrou:
Added a Faraday Response Middleware to correctly handle the Tower error cases.

In the case of a http status 400 we now pass back the body of the response
so the caller can read the messages sent back from Tower.

The original Faraday RaiseError class was simply sending back a generic message when
the response responded to `:each_key`


syncrou:
@bdunne, @mkanoor - Please review.

gmcculloug:
@syncrou Looks good.  @bdunne Please review.

bdunne:
Thanks @syncrou that looks much better

PR #27, opened 2016-04-13
Expose hash methods on job template

syncrou:


syncrou:
@bdunne, @bzwei - Please review

lib/ansible_tower_client/collection.rb
```diff
@@ -11,19 +11,19 @@ def all
     end
 
     def collection_for(paginated_result)
-      body = JSON.parse(paginated_result.body)
```
bdunne: Why are these calls changing?  It should always be JSON, right?

lib/ansible_tower_client/job_template.rb
```diff
@@ -18,12 +20,25 @@ def survey_spec
...
+      hashify(:extra_vars)
+    end
+
     def self.endpoint
       "job_templates".freeze
     end
 
     private
 
+    def self.attr_excluded?(name)
```
bdunne: I don't see this called anywhere

lib/ansible_tower_client/job_template.rb
```diff
@@ -1,5 +1,7 @@
 module AnsibleTowerClient
   class JobTemplate < BaseModel
+    EXCLUDED = [:extra_vars, :survey_spec]
```
bdunne: What are these excluded from?

spec/factories/responses.rb
```diff
@@ -19,7 +19,8 @@
     limit      { "" }
 
     trait(:description)             { sequence(:description) { |n| "description_#{n}" } }
-    trait(:extra_vars)              { extra_vars "lots of options" }
+
+    trait(:extra_vars)              { extra_vars { {:option => "lots of options"}.to_json } }
```
bdunne: `extra_vars` should remain a JSON or YAML String as it is in the api.

lib/ansible_tower_client/collection.rb
syncrou: I switched everything to YAML.safe_load for consistency since safe_load handles both types.  Though you are right - it is always going to be JSON.  If you feel that JSON.parse would be better for JSON only responses - I can make that change.

lib/ansible_tower_client/job_template.rb
syncrou: :disappointed:  - Yeah - I'll pull these out.  I thought I still needed them to allow those attributes to return only a hash.  But since we're trusting the `def launch(options = {})` options hash these are no longer needed

lib/ansible_tower_client/job_template.rb
syncrou: Same as my previous comment - pulling these out.

syncrou:
@bdunne - Pulled out the references to excluded attributes and `EXCLUDED`. I also updated all JSON only responses with JSON.parse.

bdunne:
:+1: Thanks @syncrou :+1:

PR #28, opened 2016-04-13
Handle _hash methods if survey_spec is nil or extra_vars are empty

syncrou:


syncrou:
@bdunne, @bzwei - Please review.

spec/job_template_spec.rb
```diff
@@ -23,6 +23,9 @@
   it "#extra_vars_hash" do
     obj = described_class.new(instance_double("AnsibleTowerClient::Api"), raw_instance)
     expect(obj.extra_vars_hash).to eq('option' => 'lots of options')
+    expect(obj).to receive(:extra_vars).twice.and_return('')
```
bdunne: These new lines should really be in a new `it "without extra_vars"` in a `#extra_vars_hash` context.  You should be able to build a `:response_instance` with empty `extra_vars` rather than stubbing, then `eq({})`.

bdunne:
@syncrou Can you add tests for `survey_spec_hash` too?

PR #29, opened 2016-04-20
Create find_all_by_url method

syncrou:
Extract the behavior of the Collection#all method
into find_all_by_url so we can use the same behavior
with any url passed in.

Collection#all now calls find_all_by_url.
by passing in klass.endpoint as an argument.


syncrou:
@bdunne - Please review. I created this PR to use the new Collection#find_all_by_url method in the upcoming update_all_inventory_sources PR as it makes those methods easier to read.

PR #30, opened 2016-04-20
Get extra_vars and stdout from Job

bzwei:


spec/job_spec.rb
```diff
@@ -16,4 +18,37 @@
...
+      it "returns a hashed value" do
+        obj = described_class.new(instance_double("AnsibleTowerClient::Api"), raw_instance)
+        expect(obj.extra_vars_hash).to eq('option' => 'lots of options')
+      end
+    end
+
+    describe "#extra_vars does not exist" do
+      it "returns an empty hash" do
+        obj = described_class.new(instance_double("AnsibleTowerClient::Api"), raw_instance_no_extra_vars)
+        expect(obj.extra_vars_hash).to be_a_kind_of(Hash)
```
bdunne: You could `expect(obj.extra_vars_hash).to eq({})` to reduce the number of expectations.

spec/factories/responses.rb
```diff
@@ -15,17 +15,18 @@
     type       { AnsibleTowerClient::FactoryHelper.underscore_string(klass.namespace.last) }
     name       { "#{type}-#{id}" }
     url        { "/api/v1/#{klass.endpoint}/#{id}/" }
-    related    { {"survey_spec" => "example.com/api", 'inventory' => 'inventory link'} }
+    related    { {"survey_spec" => "example.com/api", 'inventory' => 'inventory link', 'stdout' => 'example.com/api'} }
```
bdunne: Do we want to have 'stdout' on every object?

lib/ansible_tower_client/job.rb
```diff
@@ -3,6 +3,15 @@ class Job < BaseModel
...
       "jobs".freeze
     end
+
+    def extra_vars_hash
+      extra_vars.empty? ? {} : hashify(:extra_vars)
+    end
+
+    def stdout
+      out_url = related['stdout']
+      return nil unless out_url
```
bdunne: The `nil` is not required.

bdunne:
@bzwei Please rebase.

spec/factories/responses.rb
bzwei: We don't have too, but `stdout` presents in every job's related section.

PR #31, opened 2016-04-21
Add Inventory#update_all_inventory_sources

syncrou:


syncrou:
@bdunne - Please review.

syncrou:
@bdunne - Any thoughts on this PR?

lib/ansible_tower_client/inventory.rb
```diff
@@ -4,6 +4,16 @@ def self.endpoint
       "inventories".freeze
     end
 
+    def inventory_sources
+      api.inventories.find_all_by_url(related['inventory_sources'])
```
bdunne: `api.inventories.find(inventory_sources)` seems misleading.  I think you want to do something similar to the relations on other classes.

lib/ansible_tower_client/inventory_source.rb
```diff
@@ -0,0 +1,21 @@
+module AnsibleTowerClient
+  class InventorySource < BaseModel
+    def self.endpoint
+      "inventory_sources".freeze
+    end
+
+    def can_update?
+      response = api.get("#{url}update").body
```
bdunne: Is `url` guaranteed to end with `/`?  Would `File.join(url, "update")` be safer?

bdunne: Actually, you can get the full path via `related["update"]`

lib/ansible_tower_client/inventory_source.rb
```diff
@@ -0,0 +1,21 @@
+module AnsibleTowerClient
+  class InventorySource < BaseModel
+    def self.endpoint
+      "inventory_sources".freeze
+    end
+
+    def can_update?
```
bdunne: Would `updatable?` be a better name?  <-- Just a thought, not saying you have to change it if you don't like it

lib/ansible_tower_client/inventory_source.rb
```diff
@@ -0,0 +1,21 @@
...
+      response = api.get("#{url}update").body
+
+      updatable = JSON.parse(response)
+      updatable['can_update']
+    end
+
+    def update
+      response = api.post("#{url}update").body
```
bdunne: Same comment about the path

spec/inventory_spec.rb
```diff
@@ -17,4 +18,40 @@
...
+      expect(inventory_sources).to be_a Array
+      expect(inventory_sources.count).to eq 1
+    end
+  end
+
+  describe "#update_all_inventory_sources" do
+    let(:can_update_true)   { {'can_update' => true}  }
+    let(:can_update_false)  { {'can_update' => false} }
+    let(:post_result_body) { {:inventory_update => 1} }
+    let(:inventory_sources) do
```
bdunne: Shouldn't this be singular?

bdunne:
@syncrou Sorry for the delay... must have missed the e-mail.

syncrou:
@bdunne - No problem. I'll get those changes in.

lib/ansible_tower_client/inventory_source.rb
syncrou: @bdunne - I liked how can_update? read on https://github.com/syncrou/ansible_tower_client/blob/a5d91334e8ee883f008ee8a4a1462ca86f6501a3/lib/ansible_tower_client/inventory.rb?#L13 in terms of thestatement:  If inventory source can update?

lib/ansible_tower_client/inventory_source.rb
syncrou: I'll make this change.  Per the tower docs every endpoint ends with "/" - but I like seeing `related["update"]` in the code - makes it easier to read what we're calling.

spec/inventory_spec.rb
syncrou: No - That builds out the method `inventory_sources` in `inventory.rb`

bdunne: I guess I was confused because it is plural, but returns a single object.  it then gets wrapped in an array [here](https://github.com/ManageIQ/ansible_tower_client/pull/31/files#diff-7fedb3c455356ad23796abdd4747cd91R41).  I guess I was expecting it to be singular or return an array.

lib/ansible_tower_client/inventory_source.rb
bdunne: Since the field in the Tower API is "can_update", we should probably leave it as you wrote it.

spec/inventory_spec.rb
```diff
@@ -17,4 +18,40 @@
...
+      [described_class.new(instance_double("AnsibleTowerClient::Api"), inventory_source)]
```
syncrou: @bdunne - I return an array now - Which I think clears up the naming confusion.

lib/ansible_tower_client/inventory.rb
syncrou: @bdunne - Per your comment above - I looked into this - but realized for `root_groups` we're doing the same lookup - should we rename `find_all_by_url`? I like the utility it provides with only a url as an argument. 

Plus `find_all_by_url` is already calling `collection_for` but without all the nested arguments.

PR #32, opened 2016-04-29
Change Collection to use lazy enumerators

bdunne:
### Better handle paging
Rewrite collections to return lazy Enumerators instead of Arrays

Break out:
- response parsing
- result set parsing
- building objects out of results
- fetching pages

Merge #find_all_by_url and #collection_for after refactoring

Lazy Enumerators allow for quicker initial response since a page is only requested and its objects created when needed rather than fetching all pages at once then creating objects from each result.

Make 'klass' optional since it is only required for #find

Update Collection specs accordingly

##### Solved:
- The old solution is only going to get slower as the amount of collection items increases
- Long initial delay before chained methods (collect, each) can start processing
- The old solution will fetch all records even if we only care about the first N

##### Created?
- `all` returns an enumerable instead of an Array

_old solution = Fetching all pages, then turning everything in the collection into objects_

### DIFF
--- a/lib/ansible_tower_client/collection.rb
+++ b/lib/ansible_tower_client/collection.rb
@@ -1,7 +1,7 @@
 module AnsibleTowerClient
   class Collection
     attr_reader :api, :klass
-    def initialize(api, klass)
+    def initialize(api, klass = nil)
       @api   = api
       @klass = klass
     end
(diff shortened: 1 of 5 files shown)


bdunne:
@syncrou Please review
cc @gmcculloug @bzwei

lib/ansible_tower_client/collection.rb
```diff
@@ -11,37 +11,47 @@ def all
     end
```
syncrou: @bdunne - do we need to add a `.map {|x| x}` (or `.collect..`) to the `all` method?  I'm just getting an Enumerator class back.

bdunne: @syncrou The refresh code on ManageIQ will call `.each` on the collections (now enumerators). https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/ansible_tower/configuration_manager/refresh_parser.rb#L50

bdunne: collecting or mapping the enumerator will force it and immediately load everything.

syncrou: Cool - I was more concerned about the refresh code.  Thanks

spec/support/shared_examples/collection_methods.rb
```diff
@@ -1,22 +1,21 @@
 shared_examples_for "Collection Methods" do
   it ".all returns a collection of objects" do
-    expect(api).to receive(:get).and_return(instance_double("Faraday::Result", :body => raw_collection.to_json))
+    expect(collection).to receive(:find_all_by_url).with(described_class.endpoint)
 
-    obj_collection = collection.all
-    expect(obj_collection).to        be_a Array
-    expect(obj_collection.length).to eq(2)
-    expect(obj_collection.first).to  be_a described_class
+    collection.all
```
syncrou: Should we call `.map` on `collection.all` and re add `expect(collection.length).to eq(2)` etc to better match the test name?  Or just change the name to reflect that an `Enumerator` is returned?

lib/ansible_tower_client/collection.rb
syncrou: @bdunne - I guess the more I think about this -  I would expect the all method to return a collection and in doing so would more closely match the `.all` method from active_record. 

I get now that we can act on the `Enumerator` class (`.each`, `.map` etc) - but in that case maybe we should rename `all` to something else?

lib/ansible_tower_client/collection.rb
```diff
@@ -11,37 +11,47 @@ def all
...
-    end
-
-    def collection_for(paginated_result)
-      body = JSON.parse(paginated_result.body)
-      results = body["results"]
-      loop do
-        break if body["next"].nil?
-        body = JSON.parse(api.get(body["next"]).body)
-        results += body["results"]
+      Enumerator.new do |yielder|
```
syncrou: I really like this!

syncrou:
👍

bdunne:
@syncrou As suggested, I updated the test title based on what it is actually testing

spec/support/shared_examples/collection_methods.rb
```diff
@@ -1,22 +1,21 @@
 shared_examples_for "Collection Methods" do
-  it ".all returns a collection of objects" do
-    expect(api).to receive(:get).and_return(instance_double("Faraday::Result", :body => raw_collection.to_json))
+  it ".all returns an enumerator that will delay load a collection" do
```
syncrou: 👍

bdunne:
@gmcculloug Please review

PR #35, opened 2016-07-07
Limit ActiveSupport version if Ruby version is less than 2.2.2

bdunne:
Rails 5 isn't compatible with older rubies.
Upgrade more_core_extensions to provide `blank?`

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,2 +1,3 @@
 --format documentation
 --color
+--require spec_helper
--- a/ansible_tower_client.gemspec
+++ b/ansible_tower_client.gemspec
@@ -19,10 +19,13 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+  # HACK: Rails 5 dropped support for Ruby < 2.2.2
+  active_support_version = "< 5" if Gem::Version.new(RUBY_VERSION) <= Gem::Version.new("2.2.2")
+  spec.add_runtime_dependency "activesupport", active_support_version
+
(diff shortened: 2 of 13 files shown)


bdunne:
@Fryguy Please review

bdunne:
@syncrou Please review

ansible_tower_client.gemspec
```diff
@@ -19,10 +19,13 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+  # HACK: Rails 5 dropped support for Ruby < 2.2.2
+  active_support_version = "< 5" if Gem::Version.new(RUBY_VERSION) <= Gem::Version.new("2.2.2")
+  spec.add_runtime_dependency "activesupport", active_support_version
```
syncrou: My only problem with this - It would be great if we could pull out active_support all together - granted if we rolled our own version of `camelize` - which may not be trivial.

Outside of that - I would like to pull out the term #HACK: in the comment.  The change will probably live there for a long time - and #HACK: seems to indicate this change is extremely temporary.

.rspec
```diff
@@ -1,2 +1,3 @@
 --format documentation
 --color
+--require spec_helper
```
syncrou: Forgot we had .rspec as part of the gem - I like this change a lot (and the subsequent deletes in the rspec files.)  🍕

ansible_tower_client.gemspec
bdunne: I agree, it would be nice to remove the dependency on `activesupport`, but right now we use a few very complicated parts of `activesupport/inflector`

PR #44, opened 2016-08-29
Dynamically select classes based on the API version

bdunne:
We have a hard-coded patch for the V2 API on `JobTemplate#launch` that isn't needed for V3.  Instead, it would be better to start with a clean base model that defines the expected API interaction, then only patch the methods necessary for the version of the API we are interacting with.


bdunne:
@Fryguy @syncrou @gmcculloug Please review.

I'm trying to verify the change on a Tower v3 system, but having trouble getting Tower configured properly.  I'll remove the WIP once I can verify it.

spec/ad_hoc_command_spec.rb
```diff
@@ -2,6 +2,7 @@
   let(:url)                 { "example.com/api/v1/ad_hoc_command_update/10" }
   let(:api)                 { AnsibleTowerClient::Api.new(instance_double("Faraday::Connection")) }
   let(:collection)          { api.ad_hoc_commands }
+  let(:described_class)     { api.ad_hoc_command_class }
```
Fryguy: I don't like this.  described_class has a *very* well defined meeting and I don't like overriding it in specs.  This is guaranteed to cause confusion.

bdunne: the default `described_class` won't always work.  Maybe I need to rethink the way I'm testing.

lib/ansible_tower_client/api.rb
```diff
@@ -50,5 +50,53 @@ def method_missing(method_name, *args, &block)
...
+          klass.prepend(AnsibleTowerClient::V2Patches::JobTemplateLaunch)
+        end
+      end
+    end
+
+    private
+
+    def build_class(model_name)
+      base_model_constant = AnsibleTowerClient.const_get(model_name)
+      Class.new(base_model_constant).tap do |klass|
```
Fryguy: I'm concerned this could get expensive in a similar way to how creating OpenStructs en masse used to be expensive.  I'm thinking it makes more sense to create all of these class types up front using normal inheritance, and then just picking the right one at run time, instead of dynamically creating them.  I'm pretty sure that would be possible, and not overly complex.

bdunne:
:turtle:

Fryguy:
Still don't like that described_class override.  @jrafanie Thoughts?

jrafanie:
> Still don't like that described_class override. @jrafanie Thoughts?

@Fryguy I don't know what you're referring to.  Link?

bdunne:
@Fryguy I got rid of all of the places I was doing `let(:described_class)`

spec/job_template_spec.rb
```diff
@@ -1,110 +1,35 @@
+require 'faraday' # Only because we're doubling the connection
+
 describe AnsibleTowerClient::JobTemplate do
   let(:url)                 { "example.com/api/v1/job_template_update/10" }
-  let(:api)                 { AnsibleTowerClient::Api.new(instance_double("Faraday::Connection")) }
+  let(:api)                 { AnsibleTowerClient::Api.new(connection).tap { |a| allow(a).to receive(:config).and_return(config) } }
```
syncrou: Nice!

lib/ansible_tower_client/v2/job_template_v2.rb
```diff
@@ -1,5 +1,5 @@
 module AnsibleTowerClient
-  class JobTemplate < BaseModel
+  class JobTemplateV2 < JobTemplate
```
syncrou: @bdunne - Since we're not taking advantage of the `/v2/` namespace is there any real reason to nest `JobTemplateV2` under that folder? 

On that note it would be nice to be able to use the `/v2/` namespace and not have to use the `JobTemplateV2` naming convention.

bdunne: Unfortunately the api itself isn't consistent here.  You hit `api/v1`and get the current api version.  There isn't an `api/v2` or even an `api/v3` yet.  We have to ask the config for the version and use that to decide which model version to load.  I was going to keep base models separate from version specific models and if we need to I'll create a v3 directory for any workarounds to the current version.

syncrou:
👍 Looks good to me.

PR #45, opened 2016-09-09
Move activesupport hack to Gemfile.

bdunne:
Allow tests to run with the correct version of activesupport for their
ruby version.


bdunne:
@Fryguy Please review

PR #52, opened 2017-01-06
raise the original client error if no response

syncrou:
When the API class receives a Faraday::ClientError with a nil
response, we simply bubble up that error to the caller
instead of modifying and repackaging it as a ConnectionError

### DIFF
--- a/lib/ansible_tower_client/api.rb
+++ b/lib/ansible_tower_client/api.rb
@@ -4,6 +4,8 @@ module AnsibleTowerClient
   class Api < Connection
     include Logging
 
+    DEFAULT_ERROR_MSG = "An unknown error was returned from the provider".freeze
+
     attr_reader :instance
     def initialize(connection)
       @instance = connection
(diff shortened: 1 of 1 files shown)


syncrou:
@bdunne - Please review

syncrou:
cc - @gmcculloug

syncrou:
@mkanoor - Please review

lib/ansible_tower_client/api.rb
```diff
@@ -4,6 +4,8 @@ module AnsibleTowerClient
   class Api < Connection
     include Logging
 
+    DEFAULT_ERROR_MSG = "An unknown error was returned from the provider"
```
bdunne: If it is necessary to make this a constant, please freeze the string.

syncrou: I don't know if it's necessary - I generally like to see default text in a constant for readability.

I'll freeze it.  Thanks!

PR #53, opened 2017-01-06
Add create method to the base_model

syncrou:
Opens create to all subclasses
The InventorySource class does not support POST's per the API
Overriding the self.create method to save the lookup

### DIFF
--- a/lib/ansible_tower_client/base_model.rb
+++ b/lib/ansible_tower_client/base_model.rb
@@ -37,6 +37,11 @@ module AnsibleTowerClient
       super(raw_hash)
     end
 
+    def self.create(api, attributes)
+      response = api.post("#{endpoint}/", attributes).body
+      new(api, JSON.parse(response))
+    end
+
     def hashify(attribute)
       YAML.safe_load(send(attribute))
     end
--- a/spec/ad_hoc_command_spec.rb
+++ b/spec/ad_hoc_command_spec.rb
@@ -7,6 +7,7 @@ describe AnsibleTowerClient::AdHocCommand do
   let(:raw_instance)        { build(:response_instance, :klass => described_class) }
 
   include_examples "Collection Methods"
+  include_examples "Crud Methods"
 
   it "#initialize instantiates an #{described_class} from a hash" do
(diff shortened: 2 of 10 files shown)


syncrou:
@bdunne, @mkanoor please review

cc @jameswnl

lib/ansible_tower_client/base_models/inventory_source.rb
```diff
@@ -4,6 +4,10 @@ def self.endpoint
       "inventory_sources".freeze
     end
 
+    def self.create(_api, _attributes)
+      "POST not allowed"
```
Fryguy: Should this raise an exception or does it just return the String?

syncrou: @Fryguy - it should probably raise an error - Making a direct call to the API does.  I'll get that changed.

lib/ansible_tower_client/base_model.rb
```diff
@@ -37,6 +37,11 @@ def initialize(api, json_or_hash)
       super(raw_hash)
     end
 
+    def self.create(api, attributes)
+      response = api.post("#{self.endpoint}/", attributes).body
```
bdunne: The bot should complain about this redundant `self`.  The config was just added, you may need to rebase the PR.

lib/ansible_tower_client/base_models/inventory_source.rb
bdunne: Why can't you do this?

syncrou: @bdunne - POST isn't allowed on that endpoint:

screen shot 2017-01-09 at 9 12 36 am

Would we rather allow that class to call `create` and depend on the API to raise the error, or save the lookup and raise the same error ourselves from the gem?

syncrou:
@bdunne, @Fryguy - I pulled out the inventory_sources `def self.create` method for now.  I'm not sure how tower allows for creation of inventory sources, and I'm thinking that method should be overwritten once that knowledge is known.  For now I'll rely on the raise from the API request.

bdunne:
Kicking rubocop...

bdunne:
image

PR #54, opened 2017-01-09
add activity stream as an api method

durandom:
also adds ability to pass down options to collections

### DIFF
--- a/spec/activity_stream_spec.rb
+++ b/spec/activity_stream_spec.rb
@@ -1,0 +1,9 @@
+describe AnsibleTowerClient::ActivityStream do
+  let(:url)                 { "example.com/api/v1/activity_stream" }
+  let(:api)                 { AnsibleTowerClient::Api.new(instance_double("Faraday::Connection")) }
+  let(:collection)          { api.activity_stream }
+  let(:raw_url_collection)  { build(:response_url_collection, :klass => described_class, :url => url) }
+  let(:raw_instance)        { build(:response_instance, :group, :klass => described_class) }
+
+  include_examples "Collection Methods"
+end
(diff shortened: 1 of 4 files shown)


durandom:
@miq-bot assign @bdunne @syncrou 

 @bdunne @syncrou please review

durandom:
@syncrou extracted to  https://github.com/ansible/ansible_tower_client_ruby/pull/57

spec/activity_stream_spec.rb
```diff
@@ -0,0 +1,9 @@
+describe AnsibleTowerClient::ActivityStream do
+  let(:url)                 { "example.com/api/v1/activity_stream" }
+  let(:api)                 { AnsibleTowerClient::Api.new(instance_double("Faraday::Connection")) }
+  let(:collection)          { api.activity_stream }
+  let(:raw_url_collection)  { build(:response_url_collection, :klass => described_class, :url => url) }
+  let(:raw_instance)        { build(:response_instance, :group, :klass => described_class) }
+
+  include_examples "Collection Methods"
```
syncrou: The shared example "Crud Methods" could be included in here.

durandom: I'm afraid this endpoint only allows GET, HEAD, OPTIONS per 
https://ansible_tower3/api/v1/activity_stream/

bdunne: We can split the "Crud Methods" shared example into multiple shared examples (Create, Update, Delete, etc.).

durandom: https://github.com/ansible/ansible_tower_client_ruby/blob/master/spec/support/shared_examples/crud_methods.rb only has an example for 'POST' 

Do you want shared examples for read in this PR? I'd rather like to do it in follow up, as I need this to move further for the event catcher in miq

bdunne: Since we are not 'POST'ing to this endpoint, we can deal with it later.

syncrou:
👍 LGTM

PR #55, opened 2017-01-09
Added update and delete to the base_model

syncrou:
`update`: updates the current instance and applies the updated attributes to the caller.
`generate_writer?` was enabled for all subclasses to allow the caller to be updated.

`delete`: deletes the record and returns a copy of the original record.

### DIFF
--- a/lib/ansible_tower_client/base_model.rb
+++ b/lib/ansible_tower_client/base_model.rb
@@ -42,6 +42,19 @@ module AnsibleTowerClient
       new(api, JSON.parse(response))
     end
 
+    def update(attributes)
+      @api.patch(url, attributes.to_json)
+      attributes.each do |method_name, value|
+        self.send("#{method_name}=",value)
+      end
+      self
+    end
+
+    def delete
+      @api.delete(url)
+      self
+    end
+
     def hashify(attribute)
       YAML.safe_load(send(attribute))
     end
@@ -78,7 +91,7 @@ module AnsibleTowerClient
       end
 
       def generate_writer?
-        false
+        true
       end
     end
   end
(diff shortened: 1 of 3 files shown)


syncrou:
@bdunne, @Fryguy - please review.

-cc @jameswnl, @gmcculloug

syncrou:
@miq-bot assign @bdunne

lib/ansible_tower_client/base_model.rb
```diff
@@ -78,7 +91,7 @@ def key_to_attribute(key)
       end
 
       def generate_writer?
-        false
+        true
```
bdunne: I don't really like this idea.  This will generate attribute writer methods.  If those methods are publicly visible, users may get the wrong idea and expect the writer methods to update the API with the provided information, but it will just update the in memory AnsibleTowerClient object.

spec/support/shared_examples/crud_methods.rb
```diff
@@ -1,6 +1,18 @@
 shared_examples_for "Crud Methods" do
-  it ".create posts to the api and returns the new instance" do
+  it "described_class.create posts to the api and returns the new instance" do
```
bdunne: Did you mean to interpolate the `described_class`?

syncrou: I didn't - but I also didn't think about how much easier the test name will be to read if I did.
I'll get that changed.

lib/ansible_tower_client/base_model.rb
syncrou: What if I added a `save` method so in the case a user does intend to modify the in memory object, they have a way to persist their changes in an AR fashion?

bdunne:
@syncrou Based on our discussion yesterday, you're going to make this feel more like ActiveRecord, correct?
Summary:
- update -> update_attributes!
- Add update_attributes
- delete -> destroy
- Keep the `def generate_writer?; true; end` (You could actually delete the method from BashModel since it is already `true` in HashModel)
- Add save
- Add save!

PR #56, opened 2017-01-10
Expose the /projects endpoint

syncrou:
https://www.pivotaltracker.com/story/show/137293703

### DIFF
--- a/spec/project_spec.rb
+++ b/spec/project_spec.rb
@@ -1,0 +1,24 @@
+require 'json'
+
+describe AnsibleTowerClient::Project do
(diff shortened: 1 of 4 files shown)


syncrou:
@miq-bot assign @bdunne

spec/project_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'json'
+
+describe AnsibleTowerClient::Project do
+  let(:url)                 { "example.com/api/v1/project_update/10" }
```
bdunne: Shouldn't the URL be `"example.com/api/v1/projects/10"`?

PR #57, opened 2017-01-11
allow passing of GET options to .all in collections

durandom:
adds ability to pass down options to collections

```ruby
filter = {
 :order_by      => 'timestamp',
 :timestamp__gt => '2016-08-02T17:56:37.212874Z'
}
connection.api.activity_stream.all(filter)
```

see sorting, searching and filtering in the tower api docs https://docs.ansible.com/ansible-tower/3.0.1/html/towerapi

@miq-bot assign @syncrou

### DIFF
--- a/lib/ansible_tower_client/collection.rb
+++ b/lib/ansible_tower_client/collection.rb
@@ -6,6 +6,8 @@ module AnsibleTowerClient
       @klass = klass
     end
 
+    # @param get_options [Hash] a hash of http GET params to pass to the api request
+    #   e.g. { :order_by => 'timestamp', :name__contains => 'foo' }
     def all(get_options = nil)
       find_all_by_url(klass.endpoint, get_options)
     end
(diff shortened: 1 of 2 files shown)


bdunne:
@durandom Can you update the description with some examples of get_options?

durandom:
> @durandom Can you update the description with some examples of get_options?

sure, sorry for doing this in the first place...
I also added a bit of yardoc to the method, not sure if thats wanted though

syncrou:
👍 LGTM

lib/ansible_tower_client/collection.rb
```diff
@@ -6,17 +6,20 @@ def initialize(api, klass = nil)
       @klass = klass
     end
 
-    def all
-      find_all_by_url(klass.endpoint)
+    # @param get_options [Hash] a hash of http GET params to pass to the api request
+    #   e.g. { :order_by => 'timestamp', :name__contains => 'foo' }
+    def all(get_options = nil)
```
syncrou: @durandom - I had some concerns about passing args to the `all` method - but it seems this is the same behavior as ActiveRecord.  So I'm fine with it.

PR #66, opened 2017-02-01
Allow for alternative resource paths

bdunne:
By default Ansible Tower assumes that it is always at /api/v1/.
That may not always be the case, so this change allows us to specify an
alternative when initializing the connection.

### DIFF
--- a/lib/ansible_tower_client/api.rb
+++ b/lib/ansible_tower_client/api.rb
@@ -69,6 +69,8 @@ module AnsibleTowerClient
 
     def method_missing(method_name, *args, &block)
       if instance.respond_to?(method_name)
+        path = build_path_to_resource(args.shift)
+        args.unshift(path)
         logger.debug { "#{self.class.name} Sending <#{method_name}> with <#{args.inspect}>" }
         instance.send(method_name, *args, &block).tap do |response|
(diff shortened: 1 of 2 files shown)


bdunne:
@syncrou Please review
@carbonin Let's test with this to verify that it fixes the issue you saw with alternate paths.

carbonin:
@bdunne will this handle the paging as well as the issue with the specific collection we saw? (can't remember which one it was off the top of my head).

bdunne:
@carbonin yes, it should handle all of the cases.

syncrou:
@bdunne - Can you add the issue to this PR that @carbonin was seeing so I can have some background?

bdunne:
@syncrou The issue that @carbonin saw is that all referral urls from Ansible Tower start with `/api/v1/`.  This applies to the [next page referral](https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/collection.rb#L49) as well as the [referrals related to an instance](https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/base_models/host.rb#L4).  So, if your AnsibleTower server isn't at the root, i.e. https://server.example.com/blah/api/v1, you won't be able to follow any of the related links without this change.

Issue #68, opened 2017-02-09
initializer modifies class


durandom:
I'm a bit worried about 
```ruby
    def initialize(api, json_or_hash)
      @api = api

      raw_hash = json_or_hash.kind_of?(Hash) ? json_or_hash : JSON.parse(json_or_hash)
      self.class.send(:id_attr, *raw_hash['related'].keys) if raw_hash.key?('related')

      super(raw_hash)
    end
``` 
in [base_model](https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/base_model.rb#L39
) introduced in https://github.com/ansible/ansible_tower_client_ruby/pull/17

Not only does it modify a class whenever an instance is created, which can lead to different behaviour across different instances. It's also not thread safe, potentially leading to data corruption.

I'm also seeing inconsistent behaviour when accessing these `_id` methods

```ruby
(byebug) ap i.class
AnsibleTowerClient::JobTemplate < AnsibleTowerClient::BaseModel
(byebug) ap i.to_hash['related']
{
                        "created_by" => "/api/v1/users/1/",
                       "modified_by" => "/api/v1/users/1/",
                            "labels" => "/api/v1/job_templates/76/labels/",
                         "inventory" => "/api/v1/inventories/6/",
                           "project" => "/api/v1/projects/40/",
                        "credential" => "/api/v1/credentials/6/",
      "notification_templates_error" => "/api/v1/job_templates/76/notification_templates_error/",
    "notification_templates_success" => "/api/v1/job_templates/76/notification_templates_success/",
                              "jobs" => "/api/v1/job_templates/76/jobs/",
                      "object_roles" => "/api/v1/job_templates/76/object_roles/",
        "notification_templates_any" => "/api/v1/job_templates/76/notification_templates_any/",
                       "access_list" => "/api/v1/job_templates/76/access_list/",
                            "launch" => "/api/v1/job_templates/76/launch/",
                         "schedules" => "/api/v1/job_templates/76/schedules/",
                   "activity_stream" => "/api/v1/job_templates/76/activity_stream/",
                       "survey_spec" => "/api/v1/job_templates/76/survey_spec/"
}
(byebug) ap i.cloud_credential
nil
(byebug) ap i.cloud_credential_id
nil
(byebug) ap i.credential
*** NoMethodError Exception: undefined method `credential' for #<AnsibleTowerClient::JobTemplate:0x007f9a068e92f0>
[...]

PR #69, opened 2017-02-09
Expose credential off of JobTemplate

syncrou:
The credential attribute was simply returning the id vs
running a lookup and grabbing the AnsibleTowerClient::Credential object

### DIFF
--- a/lib/ansible_tower_client/base_models/job_template.rb
+++ b/lib/ansible_tower_client/base_models/job_template.rb
@@ -7,6 +7,12 @@ module AnsibleTowerClient
       api.jobs.find(job['job'])
     end
 
+    def credential
+      cred_id = to_hash['credential']
+      return nil unless cred_id
+      api.credentials.find(cred_id)
+    end
+
     def survey_spec
       spec_url = related['survey_spec']
       return nil unless spec_url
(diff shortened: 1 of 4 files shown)


syncrou:
@miq-bot add_label bug

@miq-bot assign @bdunne

syncrou:
@bdunne, @durandom - We weren't actually doing a lookup on credential as would be expected. I added the lookup in a similar pattern to how we grab the `survey_spec` and/or a `Job`

syncrou:
@bdunne - If everyone is happy with this solution - I can do a followup with `AnsibleTowerClient::CloudCredential` and `AnsibleTowerClient::NetworkCredential`.

lib/ansible_tower_client/base_models/job_template.rb
```diff
@@ -7,6 +7,12 @@ def launch(options = {})
       api.jobs.find(job['job'])
     end
 
+    def credential
+      cred_id = to_hash['credential']
+      return nil unless cred_id
+      api.credentials.find(cred_id)
```
gmcculloug: Last two lines can be combined: `api.credentials.find(cred_id) if cred_id `

durandom:
@syncrou in miq ansible refresh I am using the `_id` fields https://github.com/ManageIQ/manageiq/pull/13834/files#diff-46a11c7f5485160ff1f90f895e64e765R37 - so I dont need this magic

But while at it, maybe [the second part of this issue](https://github.com/ansible/ansible_tower_client_ruby/issues/68#issuecomment-278728495) is related?

PR #71, opened 2017-02-15
All client errors should derive from AnsibleTowerClient::ClientError

### LINKED ISSUE #70: Add error handling to the verify_credentials method
When doing a simple credential validation the caller of `AnsibleTowerClient::Connection.api.verify_credentials` are having to catch all sorts of implementation specific error classes for the case where the credentials may be valid, but we fail for some other reason (connection failures, SSL failures, etc.).

This often looks something like:
```ruby
def valid?
  connection = AnsibleTowerClient::Connection.new(
                 :base_url => URI::HTTP.build(:host => "host.example.com", :path => "/api/v1", :port => 80).to_s,
                 :username => "admin",
                 :password => "supersecret"
               )
  begin
    connection.api.verify_credentials
  rescue Faraday::ConnectionFailed, Faraday::SSLError, AnsibleTowerClient::ConnectionError, AnsibleTowerClient::ClientError
    return false
  end
  true
end
```

It would be great to have just a single error so the caller doesn't need to know to catch Faraday errors, or even to have this kind of behavior encapsulated into a `#credentials_valid?` method which would just return true or false.

bdunne:
Fixes #70

### DIFF
--- a/lib/ansible_tower_client/api.rb
+++ b/lib/ansible_tower_client/api.rb
@@ -78,8 +78,10 @@ module AnsibleTowerClient
       else
         super
       end
-    rescue Faraday::ConnectionFailed, Faraday::SSLError => err
-      raise
+    rescue Faraday::ConnectionFailed => err
+      raise AnsibleTowerClient::ConnectionError, err
+    rescue Faraday::SSLError => err
+      raise AnsibleTowerClient::SSLError, err
     rescue Faraday::ClientError => err
       raise if err.response.nil?
       response = err.response
--- a/lib/ansible_tower_client/exception.rb
+++ b/lib/ansible_tower_client/exception.rb
@@ -1,7 +1,10 @@
 module AnsibleTowerClient
-  class Error < Exception; end
-  class ClientError < Error; end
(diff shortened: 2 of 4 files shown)


bdunne:
@syncrou @carbonin Please review

syncrou:
@bdunne - It looks good - any thoughts on the second request in https://github.com/ansible/ansible_tower_client_ruby/issues/70 in regards to wrapping the behavior into a `credentials_valid?` method?

bdunne:
@syncrou I think the second part was just an example

gmcculloug:
@bdunne Does this require additional changes in the ManageIQ/manageiq repo to address existing  `Faraday` rescues like: [app/models/manageiq/providers/ansible_tower/automation_manager/job.rb#L37](https://github.com/ManageIQ/manageiq/blob/9e2b41dd6d381c458eb29907b9ed9f6af9887091/app/models/manageiq/providers/ansible_tower/automation_manager/job.rb#L37)

bdunne:
@gmcculloug Yes, it does.  I'll make a followup PR to ManageIQ.

PR #77, opened 2017-02-24
Expose plays, JobEvent

syncrou:
1. Expose `job_plays` off of `AnsibleTowerClient::Job`
2. Add `job_events` as a Collection
3. Add `AnsibleTowerClient::JobEvent` 

Testing Notes

1. Lookup a Job
`job = connection.api.jobs.find n`

2. Call job_plays off of said job
`job.job_plays.first`

3. Returned object will be a `AnsibleTowerClient::JobEvent`

### DIFF
--- a/spec/job_spec.rb
+++ b/spec/job_spec.rb
@@ -3,8 +3,8 @@ describe AnsibleTowerClient::Job do
   let(:api)            { AnsibleTowerClient::Api.new(instance_double("Faraday::Connection")) }
   let(:collection)     { api.jobs }
   let(:raw_collection) { build(:response_collection, :klass => described_class) }
-  let(:raw_url_collection)  { build(:response_url_collection, :klass => described_class, :url => url) }
-  let(:raw_instance)   { build(:response_instance, :job, :klass => described_class) }
+  let(:raw_url_collection) { build(:response_url_collection, :klass => described_class, :url => url) }
+  let(:raw_instance) { build(:response_instance, :job, :klass => described_class) }
   let(:raw_instance_no_extra_vars) { build(:response_instance, :job_template, :klass => described_class, :extra_vars => '') }
   let(:raw_instance_no_output)     { build(:response_instance, :job_template, :klass => described_class, :related => {}) }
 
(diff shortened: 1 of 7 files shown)


syncrou:
@miq-bot assign @gmcculloug 

cc - @bzwei

gmcculloug:
@bdunne Please review.

spec/job_spec.rb
```diff
@@ -37,6 +37,21 @@
     end
   end
 
+  context '#job_plays' do
+    let(:url)            { "example.com/api/v1/job_plays" }
+    let(:job_collection) { build(:response_collection, :klass => described_class) }
+    let(:job_events)     { build(:response_url_collection, :klass => AnsibleTowerClient::JobEvent, :url => url) }
+    it "returns a collection of AnsibleTowerClient::JobEvents" do
```
bzwei: Why `job_plays` returns a collection of `JobEvents`? Shouldn't them be `JobPlays`?

syncrou: @bzwei - Yes - it should be - but interestingly enough - I'm getting job_events back - so I'm digging into why.  

I was thinking the endpoint /jobs/5/job_plays/2 would actually return a JobEvent - it isn't - the related field points to job_events.  Trying to see why our Gem is returning a different type.

syncrou: @bzwei -Not sure how we should handle this - The Gem is doing the right thing - `job_plays` return data of `type` = `job_event` Here is a sample directly from the Tower API:

```
GET /api/v1/jobs/56/job_plays/
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept
X-API-Time: 0.040s




{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "play": "Hello World Sample",
            "failed_count": 0,
            "changed_count": 0,
            "started": "2016-11-29T21:34:42.175Z",
            "changed": false,
            "related": {
                "job_event": "/api/v1/job_events/151/"
            },
            "skipped_count": 0,
            "failed": false,
            "ok_count": 2,
            "unreachable_count": 0,
            "type": "job_event",
            "id": 151
        }
    ]
}
```

bzwei: Why did you say "The Gem is doing the right thing"? Your example shows it is job_play, not job_event.

syncrou: With the current discussion in regards to a `JobPlay` class for results off of the `/job_plays` endpoint - I suggest we merge this as is and I will build a new PR to address the type override.

bdunne: :+1: Sounds good.

PR #78, opened 2017-02-28
Override type when appropriate

syncrou:
The Tower API returns `job_event` summary types when querying the `/job_plays/` endpoint. 

In this PR we create a mechanism that allows us to force a type (in this case `AnsibleTowerClient::JobPlay` when we don't want to use the `type` column passed back by the Tower provider.

### DIFF
--- a/lib/ansible_tower_client.rb
+++ b/lib/ansible_tower_client.rb
@@ -21,6 +21,7 @@ require "ansible_tower_client/base_models/inventory_source"
 require "ansible_tower_client/base_models/inventory_update"
 require "ansible_tower_client/base_models/job"
 require "ansible_tower_client/base_models/job_event"
+require "ansible_tower_client/base_models/job_play"
 require "ansible_tower_client/base_models/job_template"
 require "ansible_tower_client/base_models/project"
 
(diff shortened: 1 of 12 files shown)


syncrou:
@miq-bot assign @bdunne 

-cc @bzwei @gmcculloug

gmcculloug:
@bzwei Please review

bzwei:
@syncrou I think we can completely remove the dependency on the `type` thus `class_from_type`. https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/collection.rb#L4 should always demand argument `klass` without default to `nil`.

You can see the first level list methods defined in `Api` such as `jobs`, `projects`, they all pass in the `klass` argument.  For second level resources you should also pass in the `klass` argument, like what you just did for job/job_plays.

syncrou:
@bzwei - A quick overview of `/unified_job_templates/` shows a collection of different types returned. Pulling out `class_from_type` and forcing that we set `klass` for every collection will no longer allow the gem to correctly handle the multi-typed collection returned in the above case.

PR #80, opened 2017-03-08
Allow formatted stdout, default to plain text

bzwei:
API already supports differently formatted stdout.


gmcculloug:
@syncrou Please review.

PR #85, opened 2017-04-12
Add job_events to job class and remove job_plays

bzwei:
Job play is a one type of job events. It is up to the application to make filtering through the `options` to get job plays from job events.

Job plays are thus deprecated and removed from the gem.

https://www.pivotaltracker.com/story/show/143606153


syncrou:
@bzwei - I believe we can pull this out too - https://github.com/ansible/ansible_tower_client_ruby/blob/master/spec/factories/responses.rb#L48

syncrou:
@bzwei - One more question - Should this be pulled out too? - https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/collection.rb#L68-L70 - We no longer need to override types like we had to for `job_plays`

bzwei:
@syncrou I think we can leave it as now in case we need similar handling for other cases in the future.

bdunne:
@bzwei What else needs to override `type`?

bzwei:
@bdunne @syncrou We may not have the case that needs overriding at this moment. I think it gives us the flexibility. We came up the solution with some effects. Keeping it here so that we don't need to reinvent if the such need comes.

PR #86, opened 2017-04-13
Allow collection enumerator to be enumerated multiple times

bdunne:
Options should be cleared on subsequent calls within the loop because they are already included in the next_page url.
However each time the enumerator is enumerated it calls the initial get and needs the options passed in.


bdunne:
@bzwei @syncrou Please review
cc @durandom

durandom:
wow, this looks like a subtle bug.
So it only happens, if you call more than one `Enumerable` method on the `Enumerator`.

I wonder if we should do some caching here? 
Because this means it also fires as many api requests as methods are called _and_ even worse, the api requests can have different responses

PR #87, opened 2017-04-13
Add BaseModel#reload

bdunne:
@durandom @syncrou Please review

### DIFF
--- a/lib/ansible_tower_client/base_model.rb
+++ b/lib/ansible_tower_client/base_model.rb
@@ -157,6 +157,10 @@ module AnsibleTowerClient
       YAML.safe_load(send(attribute))
     end
 
+    def reload
+      api.send(self.class.endpoint).find(id)
+    end
+
     private
 
     # convert a hash to an instance of nested model class
--- a/spec/activity_stream_spec.rb
+++ b/spec/activity_stream_spec.rb
@@ -7,4 +7,5 @@ describe AnsibleTowerClient::ActivityStream do
 
   include_examples "Collection Methods"
   include_examples "Api Methods"
+  include_examples "Instance#reload"
 end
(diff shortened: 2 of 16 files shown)


lib/ansible_tower_client/base_model.rb
```diff
@@ -157,6 +157,10 @@ def hashify(attribute)
       YAML.safe_load(send(attribute))
     end
 
+    def reload
```
syncrou: @bdunne - The call works, but using the following test - I'm not getting the caller updated as I would have expected:

```
project = conn.api.projects.all.first
2.3.0 :008 > project.description
 => "A great demo"
2.3.0 :009 > project.description = "blah"
 => "blah"
2.3.0 :011 > project.reload
<output of the called object>
2.3.0 :013 > project.description
 => "blah"
```
I would have expected `project.description` to equal 'A great demo'

I had that same issue with `update_attributes` and had to re call each method to update them [inline](https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/base_model.rb#L84-#L89)

durandom: good catch @syncrou 

rails AR also replaces the current attributes. 👇 
would it be enough to replace `@raw_hash` ?

```
def reload(options = nil)
  self.class.connection.clear_query_cache

  fresh_object =
    if options && options[:lock]
      self.class.unscoped { self.class.lock(options[:lock]).find(id) }
    else
      self.class.unscoped { self.class.find(id) }
    end

  @attributes = fresh_object.instance_variable_get('@attributes')
  @new_record = false
  self
end
```

bdunne:
Abandoning, I'll come back to it if needed.

PR #88, opened 2017-04-19
Collection tests

bdunne:
I'm trying to improve the way we test the gem and reduce the complexity and stubbing.  This is just one step in that direction.  There will be more followup PRs.


bdunne:
@syncrou Please review

spec/activity_stream_spec.rb
```diff
@@ -5,6 +5,5 @@
   let(:raw_url_collection)  { build(:response_url_collection, :klass => described_class, :url => url) }
```
syncrou: @bdunne - Are we using `:raw_url_collection` any more due to your change?

bdunne: Nope, that's coming up in a followup PR

PR #90, opened 2017-04-20
Remove extra cruft in activity_stream_spec.rb

bdunne:
Depends on https://github.com/ansible/ansible_tower_client_ruby/pull/88

None of these `let`s are used in the specs for this model.  Removing them to avoid confusion.


bdunne:
@syncrou Please review

PR #91, opened 2017-04-20
Add MockApi for testing

bdunne:
@syncrou I think it's about time we start testing with real data.  There's more coming, but in this PR I covered the root collection classes.

I know there is some duplication and the `get` method needs to be smarter, but I wanted to keep this relatively small so it isn't impossible to review.  Let me know what you think.

I even found and fixed a bug in a spec which proves that the previous mocking wasn't adding much value for the `#initialize` tests.

### DIFF
--- a/spec/support/mock_api.rb
+++ b/spec/support/mock_api.rb
@@ -7,9 +7,11 @@ module AnsibleTowerClient
       end
     end
 
-    def get(path)
+    def get(path, get_options = nil)
       suffix = path.split("api/v1/").last
       case suffix
+      when "ad_hoc_commands"
+        wrap_response(AdHocCommand.response)
       when "config"
         wrap_response(Config.response)
       when "me"
--- a/spec/support/mock_api.rb
+++ b/spec/support/mock_api.rb
@@ -20,6 +20,8 @@ module AnsibleTowerClient
         wrap_response(Group.response)
       when "hosts"
         wrap_response(Host.response)
+      when "inventories"
+        wrap_response(Inventory.response)
       when "me"
         wrap_response(Me.response)
       end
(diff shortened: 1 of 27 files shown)


bdunne:
I know the PR looks huge, but most of the additions are in the mock instance data.

spec/support/mock_api.rb
```diff
@@ -0,0 +1,48 @@
+module AnsibleTowerClient
```
syncrou: Do we want the `MockApi` available via the `AnsibleTowerClient` namespace? - It seems like maybe it should live under another namespace - possibly `AnsiblTowerClientTesting` or `AnsibleTowerClientMocks` - I guess in that case we could simply call the class `Api` without a collision.

bdunne: I thought about this for a couple days and feel like it's better to keep all components of the gem inside the gem namespace.

syncrou: I did more digging myself - and although I'd like to see our mocking and testing libraries under a separate namespace, we currently have the `FactoryHelper` under the `AnsibleTowerClient` namespace [here](https://github.com/ansible/ansible_tower_client_ruby/blob/master/spec/support/factory_helper.rb#L3-#L4).  So in the interest of consistency let's keep things consistent.

bdunne:
@syncrou Yeah, I agree with your concerns.  I think we have the same problems with the existing test setup, so at least it's not getting worse.  I think there is a lot of room for improvement, but I think this allows for it.

durandom:
For @syncrou valid concern:

aws-sdk-ruby does [stubbing / mocking](https://github.com/aws/aws-sdk-ruby/tree/082ea490f63225c159cce9639ccd293d2433044a/aws-sdk-core/lib/aws-sdk-core/stubbing) too. Plus they [store](https://github.com/aws/aws-sdk-ruby/tree/master/aws-sdk-core/apis/ec2/2016-11-15) their apis and example data in json files.

Maybe we could store the json repsonse of OPTIONS requests on https://ansible-tower.example.com/api/v1/projects/ in our repo and validate against that.

PR #75, opened 2017-02-23
Override RAW attributes

syncrou:
the `HashModel`  rewrites integer based attributes into attribute_id. This PR rewrites the raw values returned from
an API response into the correct '_id' style

https://www.pivotaltracker.com/story/show/140521307

PR #92, opened 2017-04-24
translate organization_id to organization

durandom:
in the same spirit as https://github.com/ansible/ansible_tower_client_ruby/pull/75

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1444398 in manageiq
and better than https://github.com/ManageIQ/manageiq/pull/14863

### DIFF
--- a/lib/ansible_tower_client/base_models/credential.rb
+++ b/lib/ansible_tower_client/base_models/credential.rb
@@ -1,4 +1,7 @@
 module AnsibleTowerClient
   class Credential < BaseModel
+    def override_raw_attributes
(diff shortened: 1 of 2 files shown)


lib/ansible_tower_client/base_models/credential.rb
```diff
@@ -1,4 +1,7 @@
 module AnsibleTowerClient
   class Credential < BaseModel
+     def override_raw_attributes
+       { :organization => :organization_id }
```
jameswnl: do we need the reverse of it?

PR #93, opened 2017-05-02
Check before update attribute of basemodel

jameswnl:
This is to fix https://bugzilla.redhat.com/show_bug.cgi?id=1445995

When caller passes in attributes that client doesn't know, even though Tower succeed in the call, client is blown up.

@miq-bot add_labels wip, bug

### DIFF
--- a/lib/ansible_tower_client/base_model.rb
+++ b/lib/ansible_tower_client/base_model.rb
@@ -82,10 +82,11 @@ module AnsibleTowerClient
     def update_attributes!(attributes)
       @api.patch(url, attributes.to_json)
       attributes.each do |method_name, value|
-        if override_raw_attributes[method_name]
-          send("#{override_raw_attributes[method_name]}=", value)
+        invoke_name = "#{override_raw_attributes[method_name] || method_name}="
+        if respond_to?(invoke_name)
+          send(invoke_name, value)
         else
-          send("#{method_name}=", value)
+          AnsibleTowerClient.logger.warn("Unknown attribute/method: #{invoke_name}. Skip updating it ...")
         end
       end
       true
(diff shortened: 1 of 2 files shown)


lib/ansible_tower_client/base_model.rb
```diff
@@ -82,10 +82,9 @@ def self.create(*args)
     def update_attributes!(attributes)
       @api.patch(url, attributes.to_json)
       attributes.each do |method_name, value|
-        if override_raw_attributes[method_name]
-          send("#{override_raw_attributes[method_name]}=", value)
-        else
-          send("#{method_name}=", value)
+        invoke_name = "#{override_raw_attributes[method_name] ? override_raw_attributes[method_name] : method_name}="
+        if respond_to?(invoke_name)
+          send(invoke_name, value)
```
blomquisg: Probably need a `else` or `rescue` here that logs a debug message saying it couldn't write to a specific attribute name.

jameswnl: Done. Also added a spec test.

lib/ansible_tower_client/base_model.rb
```diff
@@ -82,10 +82,11 @@ def self.create(*args)
     def update_attributes!(attributes)
       @api.patch(url, attributes.to_json)
       attributes.each do |method_name, value|
-        if override_raw_attributes[method_name]
-          send("#{override_raw_attributes[method_name]}=", value)
+        invoke_name = "#{override_raw_attributes[method_name] ? override_raw_attributes[method_name] : method_name}="
```
bdunne: `override_raw_attributes[method_name] ? override_raw_attributes[method_name] : method_name`
can be simplified to:
`override_raw_attributes[method_name] || method_name`

jameswnl: 👍 
done!

jameswnl:
@miq-bot remove_label wip

Issue #95, opened 2017-07-11
Model blows up when accessing a new attribute not seen by initializer


jameswnl:
Attribute accessors are only created by initializers (right?).  If we try to access a legitimate attribute (i.e. from Ansible Tower perspective) but hasn't been *seen* by initializer yet, it will blow up.

This is breaking https://travis-ci.org/ManageIQ/manageiq-providers-ansible_tower/jobs/250498451#L868.

Seems to me the simplest solution is to have `method_missing` defined to handle this.

@bdunne @syncrou what do you think?
 
This is related to https://github.com/ansible/ansible_tower_client_ruby/issues/68.

PR #96, opened 2017-07-11
method_missing for missing attr

jameswnl:
https://github.com/ansible/ansible_tower_client_ruby/issues/95

### DIFF
--- a/lib/ansible_tower_client/base_model.rb
+++ b/lib/ansible_tower_client/base_model.rb
@@ -95,11 +95,7 @@ module AnsibleTowerClient
       @api.patch(url, attributes.to_json)
       attributes.each do |method_name, value|
         invoke_name = "#{override_raw_attributes[method_name] || method_name}="
-        if respond_to?(invoke_name)
-          send(invoke_name, value)
-        else
-          AnsibleTowerClient.logger.warn("Unknown attribute/method: #{invoke_name}. Skip updating it ...")
-        end
+        send(invoke_name, value)
       end
       true
     end
--- a/lib/ansible_tower_client/base_model.rb
+++ b/lib/ansible_tower_client/base_model.rb
@@ -10,6 +10,18 @@ module AnsibleTowerClient
       base_class.to_s.split(/::/)[1].tableize.to_s.freeze
     end
 
+    def method_missing(m, *args, &_block)
+      if self.class.send(:id_attrs).include?(m.to_s)
+        super
+      else
+        self.class.send(:convert_value, m, args, self)
+      end
+    end
+
+    def respond_to_missing?(method, *)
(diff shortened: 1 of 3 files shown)


jameswnl:
@syncrou @bdunne this is a naive fix

gmcculloug:
@bzwei Please review

bzwei:
@jameswnl The idea is good. Need to add `responding_to_missings?`. Also please add a spec test.

jameswnl:
@miq-bot add_label bug

bzwei:
@bdunne The model class and methods are created while converting the first hash. If the second hash has some keys that do not appear in the first hash, you will get method not exist error if trying to access them. The fix tries to add new methods if possible.

jameswnl:
@bdunne yes please!

jameswnl:
@bdunne check my [workaround fix](https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/9/commits/80c24cecbf9653bd72e82c6b9b1533547b6dc4ac) for https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/9 to get the spec to pass

jameswnl:
@miq-bot add_label blocker

Fryguy:
I reviewed this with @jameswnl and @bdunne and what we discovered is (ignoring the current hack that's in the refresher_v2.yml) that every credential in the v2 yml does not have a domain and every credential in the v3 yaml does have a domain.  This means that when talking to the same provider, the API is consistent and this issue doesn't exist.

However, the tests don't do that.  One set of tests runs against a v2 API, and another set of tests run against v3.  Since the models are shared, they are effectively creating test contamination depending on which one loads first and defines the class.

The only way this could happen in the real world is if 2 different towers are accessed in the same process and those towers are two different API versions.  IMO, the only way to handle this is to create separate class sets based on the tower version.  Those class sets can be subclasses of a common class to allow simpler `.is_a?` calls, if that's desired.

```
Credential
|
+-- CredentialV2
|
+-- CredentialV3
```

Within a kind, we're not sure if different kinds of the same type have different payloads.  If they do, we may need to further subclass by kind.

```
Credential
|
+-- CredentialV2
|   |
|   +-- CredentialVmwareV2
|   |
|   +-- CredentialAwsV2
|
+-- CredentialV3
    |
    +-- CredentialVmwareV3
    |
    +-- CredentialAwsV3

```

Fryguy:
It seems this is sort of already done for JobTemplate: https://github.com/ansible/ansible_tower_client_ruby/blob/3af723b3eb24808bf0cb2cfded36247beff5eb57/lib/ansible_tower_client/api.rb#L155-L163 .  Perhaps this can be done strategically for Credential to get over that bug, then generically for *all* classes in a follow up.

PR #100, opened 2018-04-04
Add SystemJobTemplate#launch and the SystemJob class

carbonin:
This will allow us to trigger a system job template manually and return the created system job.

@bdunne please review

https://bugzilla.redhat.com/show_bug.cgi?id=1560478


gmcculloug:
cc @syncrou

PR #72, opened 2017-02-22
Fix error logging after changing error response classes

bdunne:
Log the error from the middleware to avoid rerescuing AnsibleTowerClient errors in Api class.

This will log:
```
[----] D, [2017-02-22T17:47:41.987732 #17232:c27108] DEBUG -- : AnsibleTowerClient::Middleware::RaiseTowerError Response:
#<struct Faraday::Env
 method=:post,
 body=
  "{\"inventory\":[\"Job Template 'inventory' is missing or undefined.\"]}",
 url=#<URI::HTTPS https://10.x.x.x/api/v1/job_templates/7/launch/>,
 request=
  #<struct Faraday::RequestOptions
   params_encoder=nil,
   proxy=nil,
   bind=nil,
   timeout=nil,
   open_timeout=nil,
   boundary=nil,
   oauth=nil>,
 request_headers=
  {"Authorization"=>"Basic YWRtaW46cGFzc3dvcmQxMjM=",
   "User-Agent"=>"Faraday v0.9.2",
   "Content-Type"=>"application/json"},
 ssl=
  #<struct Faraday::SSLOptions
   verify=false,
   ca_file=nil,
   ca_path=nil,
   verify_mode=nil,
   cert_store=nil,
   client_cert=nil,
   client_key=nil,
   certificate=nil,
   private_key=nil,
   verify_depth=nil,
   version=nil>,
 parallel_manager=nil,
 params=nil,
 response=
  #<Faraday::Response:0x0000000ff38928
   @env=#<struct Faraday::Env:...>,
   @on_complete_callbacks=[]>,
 response_headers=
  {"date"=>"Wed, 22 Feb 2017 22:47:39 GMT",
   "server"=>
    "Apache/2.4.6 (Red Hat Enterprise Linux) OpenSSL/1.0.1e-fips mod_wsgi/3.4 Python/2.7.5",
   "vary"=>"Accept,Cookie",
   "allow"=>"GET, POST, HEAD, OPTIONS",
   "x-api-time"=>"0.115s",
   "content-length"=>"67",
   "connection"=>"close",
   "content-type"=>"application/json"},
 status=400>
[----] E, [2017-02-22T17:47:41.988213 #17232:c27108] ERROR -- : AnsibleTowerClient::Middleware::RaiseTowerError Response Body:
{"inventory"=>["Job Template 'inventory' is missing or undefined."]}
```

PR #102, opened 2018-04-12
Add project last update stdout

Glutexo:
Add method to `Project` model that fetches stdout capture of its last update job. It uses `'related'` property of the Project. Returns `nil` if there is no update job for the Project.

Fetching and presenting this information is a workaround for cloning repositories from HTTPS servers with untrusted SSL certificates. Tower API does not offer an option to disable SSL check, so with the stdout user can at least see, why the cloning actually failed.

Alongside fix some subtle bugs in the `Api` class:

* Non-JSON responses were not logged correctly. This bug was hidden because with the default log level `debug` logs are skipped.
* `Faraday` was not available in the `Api` class, resulting it raising `LoadError` while catching `Faraday::*` exception.

https://bugzilla.redhat.com/show_bug.cgi?id=1513616

Required for [#72 in `manageiq-providers-ansible_tower`](https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/72).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Fixed
+- Fix debug logging of non-JSON responses [(#102)](https://github.com/ansible/ansible_tower_client_ruby/pull/102)
 
 ## [0.13.0] - 2018-04-05
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- Add Project#last_update [(#102)](https://github.com/ansible/ansible_tower_client_ruby/pull/102)
+
 ### Fixed
 - Fix debug logging of non-JSON responses [(#102)](https://github.com/ansible/ansible_tower_client_ruby/pull/102)
 
--- a/lib/ansible_tower_client/api.rb
+++ b/lib/ansible_tower_client/api.rb
@@ -96,6 +96,7 @@ module AnsibleTowerClient
     end
 
     def method_missing(method_name, *args, &block)
+      require 'faraday'
       if instance.respond_to?(method_name)
         path = build_path_to_resource(args.shift)
         args.unshift(path)
--- a/lib/ansible_tower_client/api.rb
+++ b/lib/ansible_tower_client/api.rb
@@ -102,7 +102,7 @@ module AnsibleTowerClient
         args.unshift(path)
         logger.debug { "#{self.class.name} Sending <#{method_name}> with <#{args.inspect}>" }
         instance.send(method_name, *args, &block).tap do |response|
-          logger.debug { "#{self.class.name} Response:\n#{JSON.parse(response.body).pretty_inspect}" }
+          logger.debug { "#{self.class.name} Response:\n#{log_from_response(response)}" }
         end
       else
         super
--- a/lib/ansible_tower_client/base_models/project.rb
+++ b/lib/ansible_tower_client/base_models/project.rb
@@ -17,5 +17,10 @@ module AnsibleTowerClient
 
       api.project_updates.find(update['project_update'])
     end
+
+    def last_update
+      return if related['last_update'].blank?
+      api.project_updates.find(related['last_update'])
+    end
   end
 end
--- a/spec/factories/responses.rb
+++ b/spec/factories/responses.rb
@@ -26,7 +26,14 @@ FactoryGirl.define do
     type       { AnsibleTowerClient::FactoryHelper.underscore_string(klass.namespace.last) }
     name       { "#{type}-#{id}" }
     url        { "/api/v1/endpoint/" }
-    related    { {"survey_spec" => "example.com/api", 'inventory' => 'inventory link', 'stdout' => 'example.com/api'} }
+    related do
+      {
+        "survey_spec" => "example.com/api",
+        "inventory"   => "inventory link",
+        "stdout"      => "example.com/api",
+        "last_update" => "example.com/api"
+      }
+    end
     limit      { "" }
 
     trait(:description)  { sequence(:description) { |n| "description_#{n}" } }
--- a/spec/middleware/raise_tower_error_spec.rb
+++ b/spec/middleware/raise_tower_error_spec.rb
@@ -3,10 +3,15 @@ require 'ansible_tower_client/middleware/raise_tower_error'
 
 describe AnsibleTowerClient::Middleware::RaiseTowerError do
   context "Faraday::Error" do
-    let(:env_400) { MockEnv.new('missing these attributes', 400) }
-    let(:env_402) { MockEnv.new('missing these attributes', 402) }
-    let(:env_404) { MockEnv.new('missing these attributes', 404) }
-    let(:env_407) { MockEnv.new('missing these attributes', 407) }
+    let(:message) { 'missing these attributes' }
+
+    let(:env_400) { MockEnv.new(message, 400) }
+    let(:env_402) { MockEnv.new(message, 402) }
+    let(:env_404) { MockEnv.new(message, 404) }
+    let(:env_407) { MockEnv.new(message, 407) }
+
+    let(:obj) { {'error' => message} }
+    let(:env_json) { MockEnv.new(obj.to_json, 400) }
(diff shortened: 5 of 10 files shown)


Glutexo:
Not sure whom should I ask for review. Maybe @bdunne or @jrafanie?

lib/ansible_tower_client/api.rb
```diff
@@ -1,4 +1,5 @@
 require 'pp'
+require 'faraday'
```
jrafanie: I would put a comment here:
`# We use Faraday exception class constants in method_missing below`

Or, just put the `require 'faraday'` in `method_missing` below since it's the only place (I think) that uses the `Faraday` constant.

lib/ansible_tower_client/base_models/project.rb
```diff
@@ -17,5 +17,13 @@ def update
 
       api.project_updates.find(update['project_update'])
     end
+
+    def last_update_stdout(format = 'txt')
+      return if !related.respond_to?(:last_update) or related['last_update'].nil?
```
jrafanie: I believe `or` and `||` are the same in this case... Generally, `||` is preferred except in the very specific cases where `or` might be helpful.  I still can't understand `or` most times and stay away from it.  http://www.virtuouscode.com/2010/08/02/using-and-and-or-in-ruby/

lib/ansible_tower_client/base_models/project.rb
```diff
@@ -17,5 +17,13 @@ def update
...
+
+      last_update_stdout_uri = URI("#{related['last_update']}stdout")
+      last_update_stdout_uri.query = URI.encode_www_form(format: format)
```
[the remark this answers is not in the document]
jrafanie: Sorry, I was looking at api.rb... It still feels like there should be a URI building method elsewhere.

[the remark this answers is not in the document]
Glutexo: I got inspired by https://github.com/Glutexo/ansible_tower_client_ruby/blob/afd22258e79804591a630b2ac5a2cd8d23c16c5d/lib/ansible_tower_client/base_models/job.rb#L11, only trying to avoid adding the query as a manually entered string. But yeah, we’re getting a bit un-DRY here. I’ll try to think out some an extraction to some helper class/method.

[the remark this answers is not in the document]
Glutexo: Actually, the stdout capture can be retrieved from the JSON resource too. Thus there is no need to request `/stdout?format=txt`. We can simply have

```ruby
    def last_update
      return unless related['last_update'].present?
      ProjectUpdate.new(api, api.get(related['last_update']))
    end
```

and avoid the full URI machinery.

Using the same reasoning we can get rid of `Job::stdout` as well as unnecessary. Since this is an API gem, I’d rather open an issue for it instead of blindly removing it.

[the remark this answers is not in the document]
jrafanie: I'm 👍 on less URI building.  I'm pretty new to this gem.  Does your `last_update` method return the same thing as what you have now?  Finding the most recent `project_updates` reads better to me.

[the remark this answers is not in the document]
Glutexo: The new `last_update` method returns a whole `ProjectUpdate` object, the original `last_update_stdout` returned only the stdout string. This requires an update on the MIQ provider side too, but that’s no problem.

I can avoid using `related['last_update']` and fetch the most recent update manually by `api.project_updates.all('project' => project.id, 'order_by' => '-id').first`.

Glutexo:
I fixed the `require`’s and the `respond_to?`’s. I also simplified the `last_update` method to return the whole object and not only the stdout. This avoided URI building. I stuck to using `related['last_update']` though as they are used quite a lot in this gem and this approach is IMO more readable than building a custom query.

Is it better now, @jrafanie?

lib/ansible_tower_client/base_models/project.rb
```diff
@@ -17,5 +17,10 @@ def update
 
       api.project_updates.find(update['project_update'])
     end
+
+    def last_update
+      return if related['last_update'].blank?
+      api.project_updates.find(related['last_update'])
+    end
```
jrafanie: I'm confused.  Is this method called?  I don't see it.

jrafanie: The other changes look good. 👍

Glutexo: It is not called in the API itself. The `manageiq-providers-ansible_tower` will use it.

jrafanie: I see calls to `last_updated` but not `last_update` in the https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/72 that uses this code change.  Is this right or a typo?  It looks like it's being called on a `project`.

jrafanie: I'm fine with adding new methods but have seen many times where we add code we are sure we'll need but never end up using.  I want to confirm it's actually going to be used next.

Finally, if you're comfortable with `git rebase -i`, it would be nice to squash some of the commits with "fix" or "style" in the name so each commit is standalone with a single type of change.  I don't personally like squashing all commits, but minor typos/style/fix commits are definitely worth squashing.

Glutexo: The `project.last_updated` in [ManageIQ/manageiq-providers-ansible_tower#72](https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/72) is really a typo. Thanks for finding out! I’ll fix it along with the other changes requested for that PR.

I’ll do the squashing as you suggested.

jrafanie: Don't worry about squashing commits if it becomes difficult.  It's generally good practice to commit "oops/typos/fixes" separate from logical change commits and forcing yourself to try to do this makes it easier to squash commits.  Ping me when it's ready to go.

Glutexo:
@jrafanie I squashed the commits. I also bumped the version from `0.13.0` to `0.14.0` so our `manageiq-providers-ansible_tower` “gem” can depend on the new method.

lib/ansible_tower_client/api.rb
```diff
@@ -96,12 +96,13 @@ def system_job_templates
     end
 
     def method_missing(method_name, *args, &block)
+      require 'faraday'
```
bdunne: Why `require 'faraday'` here?

spec/factories/responses.rb
```diff
@@ -26,7 +26,10 @@
     type       { AnsibleTowerClient::FactoryHelper.underscore_string(klass.namespace.last) }
     name       { "#{type}-#{id}" }
     url        { "/api/v1/endpoint/" }
-    related    { {"survey_spec" => "example.com/api", 'inventory' => 'inventory link', 'stdout' => 'example.com/api'} }
+    related    { {"survey_spec" => "example.com/api",
```
bdunne: Avoid multi-line blocks with `{}`.

spec/middleware/raise_tower_error_spec.rb
```diff
@@ -55,6 +57,14 @@
       )
       expect { error.on_complete(env_407) }.to raise_error(AnsibleTowerClient::ConnectionError)
     end
+
+    it "logs error with formatted JSON response body" do
+      response_body = { 'error' => 'missing these attributes' }.pretty_inspect
```
bdunne: Any reason not to put the string here?

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
```
bdunne: Thanks for updating the changelog!

lib/ansible_tower_client/api.rb
jrafanie: It's for the exception below 👇

bdunne: That won't actually work for `def/rescue/end` because those constants wouldn't exist at method definition time unless faraday is already required somewhere else.  I think it would work only for `def/require/begin/rescue/end/end`.

jrafanie:
@Glutexo please mention the people requesting changes if you're done implementing the changes... We don't always notice pushes to your PR branch or know if you're finished.  Thanks!

spec/factories/responses.rb
Glutexo: To be honest, I do not like this using `{}` for one-liners and `do … end` for multi-line blocks much. They behave slightly differently (not in this case) and here it breaks the nice symmetry. I understand the style standards however, so fixed.

PR #107, opened 2018-06-11
Support both API v1 and v2 at the same time

miha-plesko:
With this commit we allow user to quickly switch between v1 and v2 per API call:

```ruby
client = AnsibleTowerClient::Connection.new(
  :base_url   => 'http://localhost:54321',
  :username   => 'admin',
  :password   => 'password'
)

# Credentials from API v1
client.api(:version => 1).credentials.all.each do |c|
  pp c
end

# Credentials from API v2
client.api(:version => 2).credentials.all.each do |c|
  pp c
end
```

This way migrating ManageIQ to use v2 will be easier because we can migrate API call by API call.

Related issue: https://github.com/ManageIQ/manageiq-providers-ansible_tower/issues/96

### DIFF
--- a/lib/ansible_tower_client/connection.rb
+++ b/lib/ansible_tower_client/connection.rb
@@ -1,29 +1,67 @@
 module AnsibleTowerClient
   class Connection
-    attr_reader :connection
+    attr_reader :options
 
-    def initialize(options = nil)
+    # Options:
+    # - base_url: you have two options here:
+    #   a) pass in only scheme and hostname e.g. 'https://localhost:54321' to allow client to connect to both api v1
+    #      and v2 versions like this: `client.api(:version => 1)` and `client.api(:version => 2)`. This requires ansible
+    #      tower API being accessible directly at `https://localhost:54321/api/v1` and `https://localhost:54321/api/v2`.
+    #   b) pass in a complete api address e.g. 'https://localhost:54321/tower'. Client will then connect to the path
+    #      directly and it's your responsibility to know what version of API is there.
+    # - username
+    # - password
+    # - verify_ssl
+    def initialize(options = {})
       raise "Credentials are required" unless options[:username] && options[:password]
       raise ":base_url is required" unless options[:base_url]
-      verify_ssl = options[:verify_ssl] || OpenSSL::SSL::VERIFY_PEER
-      verify_ssl = verify_ssl == OpenSSL::SSL::VERIFY_NONE ? false : true
 
       require 'faraday'
       require 'faraday_middleware'
       require 'ansible_tower_client/middleware/raise_tower_error'
       Faraday::Response.register_middleware :raise_tower_error => -> { Middleware::RaiseTowerError }
-      @connection = Faraday.new(options[:base_url], :ssl => {:verify => verify_ssl}) do |f|
+
+      @options = {
+        :url        => options[:base_url],
+        :verify_ssl => (options[:verify_ssl] || OpenSSL::SSL::VERIFY_PEER) != OpenSSL::SSL::VERIFY_NONE,
+        :username   => options[:username],
+        :password   => options[:password]
+      }
+
+      reset
+    end
+
+    def connection(url:, username:, password:, verify_ssl: false)
+      Faraday.new(url, :ssl => {:verify => verify_ssl}) do |f|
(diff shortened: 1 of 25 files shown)


miha-plesko:
@jameswnl this PR is a result of our conversation on the linked issue. As far as I've tested, everything works as intended if we stick with API v1 for other calls and only use v2 to fetch credentials. The only difference then is that also credentials that are using custom credential types are also visible/usable.

I kindly ask for review.

miha-plesko:
Anyone? :)

bdunne:
I don't like the two different `api` and `api2` methods.  If anything, I think it would be nicer to have `client.api` default to the latest available version and optionally allow `client.api(:version => "v1")` to support using a specific version.

miha-plesko:
Thanks for the comment @bdunne . Since we're supporting only the first v2 response for now (˙client.api.credentials`), I think we'd better default to API v1 and optionally allow v2. Will come up with  code tomorrow.

bdunne:
I'd prefer to default to the latest in the gem and override it where it's used in ManageIQ

miha-plesko:
Done as you suggest @bdunne . Kindly ask for another pass of review.

lib/ansible_tower_client/api2.rb
```diff
@@ -0,0 +1,7 @@
+module AnsibleTowerClient
+  class Api2 < Api
+    def credential_class
+      @credential_class ||= AnsibleTowerClient::CredentialV2
```
bdunne: Any reason we can't switch this like the other types?
https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/api.rb#L182

lib/ansible_tower_client/connection.rb
```diff
@@ -1,29 +1,51 @@
...
+
+      @options = {
+        :uri        => URI(options[:base_url]),
+        :verify_ssl => (options[:verify_ssl] || OpenSSL::SSL::VERIFY_PEER) != OpenSSL::SSL::VERIFY_NONE,
+        :username   => options[:username],
+        :password   => options[:password]
+      }
+    end
+
+    def connection(uri:, username:, password:, api_version: 1, verify_ssl: false)
```
bdunne: I don't think api_version should have a default here, should it?

lib/ansible_tower_client/connection.rb
```diff
@@ -1,29 +1,51 @@
...
         f.adapter(Faraday.default_adapter)
-        f.basic_auth(options[:username], options[:password])
+        f.basic_auth(username, password)
       end
     end
 
-    def api
-      @api ||= Api.new(connection)
+    def api(version: 2)
+      klass = version == 1 ? Api : Api2
```
bdunne: We shouldn't need to do this

Fryguy:
@bdunne @syncrou @miha-plesko  This PR seems to have remove the connection caching altogether...was that intentional and if so, why?

Before `@connection` memoizing variable: https://github.com/ansible/ansible_tower_client_ruby/pull/107/files#diff-68ecb5d8417901430d83bc5f650e766eL15
After no memoization: https://github.com/ansible/ansible_tower_client_ruby/pull/107/files#diff-68ecb5d8417901430d83bc5f650e766eR35

miha-plesko:
@Fryguy connection caching is still there, only it was moved a bit to allow for switching between v1 and v2: https://github.com/ansible/ansible_tower_client_ruby/pull/107/files#diff-68ecb5d8417901430d83bc5f650e766eR46

Before:

```ruby
conn = AnsibleTowerClient::Connection.new(...) # "global" memoization via @connection
conn.api.credentials
conn.api.inventories
...
```

After:

```ruby
conn = AnsibleTowerClient::Connection.new(...)
conn.api(:version => 1).credentials # memoization via @api[1] for AWX API v1
conn.api(:version => 1).inventories

conn.api(:version => 2).credentials # memoization via @api[2] for AWX API v2
conn.api(:version => 2).inventories
...
```

Reason to implement memoization per api version instead globally is that URL is different for v1 and v2, hence we need to memoize the two separately:

```
http://localhost:54321/api/v1 # url for API v1
http://localhost:54321/api/v2 # url for API v2 (on the same Ansible Tower instance)
```

Please let me know if my explanation makes any sense!

PR #112, opened 2018-07-09
handling workflow_job_node not started yet

jameswnl:
When a workflow is launched, some of the the `workflow_job_nodes` won't have their job started and accessing the `job` will fail. This will fix that and return `nil`

### DIFF
--- a/lib/ansible_tower_client/base_models/workflow_job_node.rb
+++ b/lib/ansible_tower_client/base_models/workflow_job_node.rb
@@ -5,7 +5,7 @@ module AnsibleTowerClient
     end
 
     def job
-      super.nil? ? nil : api.jobs.find(job_id)
+      !respond_to?(:job_id) || job_id.nil? ? nil : api.jobs.find(job_id)
     end
   end
 end
--- a/lib/ansible_tower_client/base_models/workflow_job_node.rb
+++ b/lib/ansible_tower_client/base_models/workflow_job_node.rb
@@ -5,7 +5,7 @@ module AnsibleTowerClient
     end
 
     def job
-      job_id.nil? ? nil : api.jobs.find(job_id)
+      super.nil? ? nil : api.jobs.find(job_id)
     end
   end
 end
--- a/lib/ansible_tower_client/base_models/workflow_job_node.rb
+++ b/lib/ansible_tower_client/base_models/workflow_job_node.rb
@@ -5,7 +5,7 @@ module AnsibleTowerClient
     end
 
     def job
-      !respond_to?(:job_id) || job_id.nil? ? nil : api.jobs.find(job_id)
+      api.jobs.find(job_id) if respond_to?(:job_id) && job_id
     end
   end
 end
--- a/spec/support/mock_api/workflow_job_node.rb
+++ b/spec/support/mock_api/workflow_job_node.rb
@@ -6,6 +6,46 @@ module AnsibleTowerClient
           {
             "id": 1,
             "type": "workflow_job_node",
+            "url": "/api/v1/workflow_job_nodes/2/",
+            "related": {
+              "unified_job_template": "/api/v1/job_templates/216/",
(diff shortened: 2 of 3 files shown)


lib/ansible_tower_client/base_models/workflow_job_node.rb
```diff
@@ -5,7 +5,7 @@ def workflow_job
     end
 
     def job
-      job_id.nil? ? nil : api.jobs.find(job_id)
+      super.nil? ? nil : api.jobs.find(job_id)
```
syncrou: @jameswnl - Any way to get some specs around this change?

bdunne: This is a generic problem with the way that the BaseModel works, right?

jameswnl: right. `job_id` doesn't exist if `job/super` is not there.

gmcculloug:
@jameswnl Do you need help moving this forward?  This PR technically blocks merging https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/103 because it can raise an error if the job does not exist when it is called here https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/103/files#diff-5d6544405f28c534f089db8ee5001058R81

cc @d-m-u

lib/ansible_tower_client/base_models/workflow_job_node.rb
jameswnl: @gmcculloug @syncrou any help on how to spec on this would be greatly appreciated

syncrou: @jameswnl - I was able to write a test supporting the `super.nil?` case:
```
  it "#job returns a nil when the job_id is nil" do
    obj = api.workflow_job_nodes.all.first
    allow_any_instance_of(AnsibleTowerClient::BaseModel).to receive(:job).and_return(nil)
    obj.job_id = 1
    expect(obj.job).to be_nil
  end
```
Though, still forced to include `allow_any_instance_of` which is problematic.

I'm inclined to approve this as is, with my work on writing the tests I'm confident it will work as expected.

bdunne:
I think https://github.com/ansible/ansible_tower_client_ruby/commit/21c22ef3e62c663210a12633cd67d75fc3ed859a is a much cleaner approach.  @syncrou thoughts?

syncrou:
@bdunne - I do like the removal of the negative case on `respond_to?(:job_id)` Can you put a PR together with the changes?

jameswnl:
@bdunne I'll just steal your [line of code](https://github.com/ansible/ansible_tower_client_ruby/commit/21c22ef3e62c663210a12633cd67d75fc3ed859a) 😄

bdunne:
Ok, @jameswnl @syncrou I pushed that commit to this branch.

lib/ansible_tower_client/base_models/workflow_job_node.rb
```diff
@@ -5,7 +5,7 @@ def workflow_job
     end
 
     def job
-      job_id.nil? ? nil : api.jobs.find(job_id)
+      api.jobs.find(job_id) if respond_to?(:job_id) && job_id
```
jameswnl: contributed by @bdunne [here](https://github.com/ansible/ansible_tower_client_ruby/commit/21c22ef3e62c663210a12633cd67d75fc3ed859a#diff-bc25d095695ec0bc4c9c49a9e4ef0134)

syncrou: I think we might just want @bdunne's raw commit - the mock object isn't needed and rubocop will clear up as well.

bdunne:
I opened https://github.com/ansible/ansible_tower_client_ruby/pull/116 instead

syncrou:
@jameswnl - Can we please close in favor of #116

PR #113, opened 2018-07-16
Default API version to 1

jameswnl:
The major consumer of this gem is `manageiq-providers-ansible_tower` and which is based on v1 api. Before `manageiq-providers-ansible_tower` support v2 in, we shall keep this gem. And besides currently only the `credential_type` from v2 is supported in this gem.


jameswnl:
@bdunne Please review

jameswnl:
#117 is reverting the PRs. closing this.

PR #119, opened 2018-08-15
Check and raise if job limit option will be ignored

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1563867

"Limit" option will be ignored by Ansible Tower if job template has no "prompt limit on launch" option checked.

With this change error will be raised for that case.

### DIFF
--- a/lib/ansible_tower_client.rb
+++ b/lib/ansible_tower_client.rb
@@ -37,6 +37,11 @@ require "ansible_tower_client/v2/job_template_v2"
 
 require "more_core_extensions/all"
 require "active_support/inflector"
+require 'active_support/core_ext/object/try'
+require 'active_support/core_ext/object/blank'
+require 'active_support/core_ext/array/conversions'
+require 'active_support/core_ext/hash/slice'
+require 'active_support/core_ext/hash/except'
 
 module AnsibleTowerClient
   class << self
--- a/lib/ansible_tower_client.rb
+++ b/lib/ansible_tower_client.rb
@@ -37,11 +37,6 @@ require "ansible_tower_client/v2/job_template_v2"
 
 require "more_core_extensions/all"
 require "active_support/inflector"
-require 'active_support/core_ext/object/try'
-require 'active_support/core_ext/object/blank'
-require 'active_support/core_ext/array/conversions'
-require 'active_support/core_ext/hash/slice'
-require 'active_support/core_ext/hash/except'
 
 module AnsibleTowerClient
   class << self
--- a/lib/ansible_tower_client/base_models/job_template.rb
+++ b/lib/ansible_tower_client/base_models/job_template.rb
@@ -1,16 +1,16 @@
 module AnsibleTowerClient
   class JobTemplate < BaseModel
-    IGNORABLE_OPTIONS = {
+    REQUIRED_ATTRIBUTES_FOR_OPTIONS = {
       :job_tags   => :ask_tags_on_launch,
       :job_type   => :ask_job_type_on_launch,
       :limit      => :ask_limit_on_launch,
       :inventory  => :ask_inventory_on_launch,
       :credential => :ask_credential_on_launch,
     }.freeze
-    private_constant :IGNORABLE_OPTIONS
+    private_constant :REQUIRED_ATTRIBUTES_FOR_OPTIONS
 
     def launch(options = {})
-      check_ignorable_options(options)
+      check_required_attributes_for_options(options)
 
       launch_url = "#{url}launch/"
       response   = api.post(launch_url, options).body
@@ -40,16 +40,15 @@ module AnsibleTowerClient
 
     private
 
-    def check_ignorable_options(options)
-      ignored_options = IGNORABLE_OPTIONS.select do |option, checkmark|
+    def check_required_attributes_for_options(options)
+      ignored_options = REQUIRED_ATTRIBUTES_FOR_OPTIONS.select do |option, checkmark|
         !try(checkmark) && options.slice(option.to_sym, option.to_s).any?(&:present?)
       end.keys
 
       return if ignored_options.empty?
 
-      ignored_options.map! { |option| ":#{option}" }
-      message = ignored_options.to_sentence
-      message = ignored_options.size > 1 ? " #{message} is" : "s #{message} are"
+      message = ignored_options.map(&:inspect).to_sentence
+      message = ignored_options.size == 1 ? " #{message} is" : "s #{message} are"
       message = "Option#{message} provided but corresponding ask on launch flag has not been turn on"
       raise ArgumentError, message
     end
--- a/lib/ansible_tower_client/base_models/job_template.rb
+++ b/lib/ansible_tower_client/base_models/job_template.rb
@@ -10,7 +10,7 @@ module AnsibleTowerClient
     private_constant :REQUIRED_ATTRIBUTES_FOR_OPTIONS
 
     def launch(options = {})
-      check_required_attributes_for_options(options)
+      validate_launch_options(options)
 
       launch_url = "#{url}launch/"
       response   = api.post(launch_url, options).body
@@ -40,14 +40,14 @@ module AnsibleTowerClient
 
     private
 
-    def check_required_attributes_for_options(options)
+    def validate_launch_options(options)
       ignored_options = REQUIRED_ATTRIBUTES_FOR_OPTIONS.select do |option, checkmark|
-        !try(checkmark) && options.slice(option.to_sym, option.to_s).any?(&:present?)
+        options.slice(option.to_sym, option.to_s).any?(&:present?) && !(respond_to?(checkmark) && send(checkmark))
       end.keys
 
       return if ignored_options.empty?
 
-      message = ignored_options.map(&:inspect).to_sentence
+      message = ignored_options.map(&:inspect).join(', ')
       message = ignored_options.size == 1 ? " #{message} is" : "s #{message} are"
       message = "Option#{message} provided but corresponding ask on launch flag has not been turn on"
       raise ArgumentError, message
--- a/lib/ansible_tower_client/base_models/job_template.rb
+++ b/lib/ansible_tower_client/base_models/job_template.rb
@@ -47,10 +47,7 @@ module AnsibleTowerClient
 
       return if ignored_options.empty?
 
-      message = ignored_options.map(&:inspect).join(', ')
-      message = ignored_options.size == 1 ? " #{message} is" : "s #{message} are"
-      message = "Option#{message} provided but corresponding ask on launch flag has not been turn on"
-      raise ArgumentError, message
+      raise ArgumentError, "'PROMPT ON LAUNCH' is required for the following fields: #{ignored_options.join(', ')}"
     end
   end
 end
--- a/lib/ansible_tower_client/base_models/job_template.rb
+++ b/lib/ansible_tower_client/base_models/job_template.rb
@@ -1,9 +1,21 @@
 module AnsibleTowerClient
   class JobTemplate < BaseModel
+    IGNORABLE_OPTIONS = {
+      :job_tags   => :ask_tags_on_launch,
+      :job_type   => :ask_job_type_on_launch,
+      :limit      => :ask_limit_on_launch,
+      :inventory  => :ask_inventory_on_launch,
+      :credential => :ask_credential_on_launch,
+    }.freeze
+    private_constant :IGNORABLE_OPTIONS
+
     def launch(options = {})
+      check_ignorable_options(options)
+
       launch_url = "#{url}launch/"
       response   = api.post(launch_url, options).body
       job        = JSON.parse(response)
+
       api.jobs.find(job['job'])
     end
 
@@ -24,6 +36,22 @@ module AnsibleTowerClient
 
     def override_raw_attributes
       { :credential => :credential_id, :inventory => :inventory_id, :project => :project_id }
+    end
+
+    private
+
+    def check_ignorable_options(options)
+      ignored_options = IGNORABLE_OPTIONS.select do |option, checkmark|
+        !try(checkmark) && options.slice(option.to_sym, option.to_s).any?(&:present?)
+      end.keys
+
+      return if ignored_options.empty?
+
+      ignored_options.map! { |option| ":#{option}" }
+      message = ignored_options.to_sentence
+      message = ignored_options.size > 1 ? " #{message} is" : "s #{message} are"
+      message = "Option#{message} provided but corresponding ask on launch flag has not been turn on"
+      raise ArgumentError, message
     end
   end
 end
--- a/lib/ansible_tower_client/base_models/job_template.rb
+++ b/lib/ansible_tower_client/base_models/job_template.rb
@@ -42,7 +42,7 @@ module AnsibleTowerClient
 
     def validate_launch_options(options)
       ignored_options = REQUIRED_ATTRIBUTES_FOR_OPTIONS.select do |option, checkmark|
-        options.slice(option.to_sym, option.to_s).any?(&:present?) && !(respond_to?(checkmark) && send(checkmark))
+        options.values_at(option.to_sym, option.to_s).any?(&:present?) && !(respond_to?(checkmark) && send(checkmark))
       end.keys
 
       return if ignored_options.empty?
--- a/lib/ansible_tower_client/hash_model.rb
+++ b/lib/ansible_tower_client/hash_model.rb
@@ -71,7 +71,6 @@ module AnsibleTowerClient
 
       def add_accessor_methods(method, key)
         return if modelized_list.include?(method.to_s)
-        key = key.to_s
 
         method = "_#{method}" if instance_methods.include?(method.to_sym)
         class_eval { define_method(method) { @data[key] } }
@@ -84,7 +83,7 @@ module AnsibleTowerClient
 
     def initialize(json_or_hash)
       raw_hash = json_or_hash.kind_of?(Hash) ? json_or_hash : JSON.parse(json_or_hash)
-      @data = Hash[raw_hash.collect { |key, value| [key.to_s, self.class.send(:convert_value, key, value, self)] }]
+      @data = Hash[raw_hash.collect { |key, value| [key, self.class.send(:convert_value, key, value, self)] }]
     end
 
     def [](key)
--- a/lib/ansible_tower_client/hash_model.rb
+++ b/lib/ansible_tower_client/hash_model.rb
@@ -71,6 +71,7 @@ module AnsibleTowerClient
 
       def add_accessor_methods(method, key)
         return if modelized_list.include?(method.to_s)
+        key = key.to_s
 
         method = "_#{method}" if instance_methods.include?(method.to_sym)
         class_eval { define_method(method) { @data[key] } }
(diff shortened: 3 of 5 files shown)


AlexanderZagaynov:
@gmcculloug @bzwei please review

lib/ansible_tower_client/base_models/job_template.rb
```diff
@@ -1,9 +1,20 @@
...
+require 'active_support/core_ext/hash/indifferent_access'
+
 module AnsibleTowerClient
   class JobTemplate < BaseModel
+    class IgnorableLimit < RuntimeError; end
+
     def launch(options = {})
+      options = options.with_indifferent_access
+
+      raise IgnorableLimit if !try(:ask_limit_on_launch) && options[:limit].present?
```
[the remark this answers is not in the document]
AlexanderZagaynov: Can you write here which fields on `JobTemplate` and which their corresponding options I should check?
P.S. However, this is out of BZ scope.

spec/factories/responses.rb
```diff
@@ -35,6 +35,7 @@
       }
     end
     limit      { "" }
+    ask_limit_on_launch true
```
bdunne: This doesn't apply to everything, can you follow the trait patterns below?

spec/job_template_spec.rb
```diff
@@ -1,3 +1,5 @@
+require 'active_support/core_ext/hash/except'
```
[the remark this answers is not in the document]
bdunne: This isn't needed if you use the traits

lib/ansible_tower_client/hash_model.rb
```diff
@@ -83,7 +85,8 @@ def add_accessor_methods(method, key)
 
     def initialize(json_or_hash)
       raw_hash = json_or_hash.kind_of?(Hash) ? json_or_hash : JSON.parse(json_or_hash)
-      @data = Hash[raw_hash.collect { |key, value| [key, self.class.send(:convert_value, key, value, self)] }]
+      raw_hash = raw_hash.collect { |key, value| [key, self.class.send(:convert_value, key, value, self)] }
+      @data = ActiveSupport::HashWithIndifferentAccess[raw_hash]
```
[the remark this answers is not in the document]
bdunne: And is it needed?

lib/ansible_tower_client/base_models/job_template.rb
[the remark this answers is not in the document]
bzwei: This is the api client gem. Should not be constrained by how manageiq uses it. It is assumed `options` provides options that the tower expects. We can edit the PR title to be more accurate. Reference to the api page for accurate list of fields the launch api will take. https://<tower.url>/api/v1/job_templates/262/launch/. Examples are `limit`, `inventory`, `credential`, `verbose`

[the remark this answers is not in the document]
AlexanderZagaynov: Ok, I added the fields mentioned in the **"Inspect returned data for optionally allowed runtime data that the user should be asked for"** section of https://docs.ansible.com/ansible-tower/latest/html/towerapi/launch_jobtemplate.html to the checking.

lib/ansible_tower_client/hash_model.rb
```diff
@@ -1,3 +1,5 @@
+require 'active_support/hash_with_indifferent_access'
```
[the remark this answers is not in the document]
AlexanderZagaynov: done

lib/ansible_tower_client/hash_model.rb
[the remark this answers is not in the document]
AlexanderZagaynov: Here is the explanation:
```
$ bin/console 

2.3.0 :001 > jt = AnsibleTowerClient::JobTemplate.new(nil, :ask_limit_on_launch => true)
 => <AnsibleTowerClient::JobTemplate ask_limit_on_launch=true> 
2.3.0 :002 > jt.ask_limit_on_launch
 => true 

2.3.0 :003 > jt = AnsibleTowerClient::JobTemplate.new(nil, 'ask_limit_on_launch' => true)
 => <AnsibleTowerClient::JobTemplate ask_limit_on_launch=true> 
2.3.0 :004 > jt.ask_limit_on_launch
 => nil 
```
However, I solved that in other way in PR update.

spec/factories/responses.rb
AlexanderZagaynov: done

spec/job_template_spec.rb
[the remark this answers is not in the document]
AlexanderZagaynov: I want to be sure that key doesn't exists. And traits can only add, not remove attributes, at least I didn't find any info about that.

AlexanderZagaynov:
@gmcculloug @bzwei @bdunne PR updated

lib/ansible_tower_client/base_models/job_template.rb
```diff
@@ -25,5 +37,19 @@ def extra_vars_hash
...
+
+    def check_ignorable_options(options)
+      ignored_options = IGNORABLE_OPTIONS.select do |option, checkmark|
+        !try(checkmark) && options.slice(option.to_sym, option.to_s).any?(&:present?)
+      end.keys
+
+      return if ignored_options.empty?
+
+      ignored_options.map! { |option| ":#{option}" }
+      message = "Option#{'s' if ignored_options.size > 1} #{ignored_options.to_sentence} will be ignored"
```
bzwei: "Option will be ignored" sounds like a warning not an error. An error message like "Option is provided but corresponding ask on launch flag has not been turn on" is more helpful.

lib/ansible_tower_client/hash_model.rb
```diff
@@ -71,6 +71,7 @@ def convert_value(key, value, parent)
 
       def add_accessor_methods(method, key)
         return if modelized_list.include?(method.to_s)
+        key = key.to_s
```
bzwei: Why a symbol is required here? I am not convinced we ever need to fix `hash_model`.

lib/ansible_tower_client/hash_model.rb
```diff
@@ -83,7 +84,7 @@ def add_accessor_methods(method, key)
 
     def initialize(json_or_hash)
       raw_hash = json_or_hash.kind_of?(Hash) ? json_or_hash : JSON.parse(json_or_hash)
-      @data = Hash[raw_hash.collect { |key, value| [key, self.class.send(:convert_value, key, value, self)] }]
+      @data = Hash[raw_hash.collect { |key, value| [key.to_s, self.class.send(:convert_value, key, value, self)] }]
```
bzwei: same

AlexanderZagaynov:
@gmcculloug  @bzwei  @bdunne  PR updated

lib/ansible_tower_client/hash_model.rb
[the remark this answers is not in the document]
bzwei: `JobTemplate` object is always created from the JSON response of the API request. They key is always a string not symbol. 

My test is the opposite
```
irb(main):011:0> jt = AnsibleTowerClient::JobTemplate.new(nil, 'ask_limit_on_launch' => true)
=> <AnsibleTowerClient::JobTemplate ask_limit_on_launch=true>
irb(main):012:0> jt.ask_limit_on_launch
=> true
irb(main):013:0> jt = AnsibleTowerClient::JobTemplate.new(nil, :ask_limit_on_launch => true)
=> <AnsibleTowerClient::JobTemplate ask_limit_on_launch=true>
irb(main):014:0> jt.ask_limit_on_launch
=> nil
```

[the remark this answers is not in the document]
AlexanderZagaynov: I did that for reason. Not just because I wanted to.
I have no idea, what happens with your test.
My tests are failing without this change. 
Even the specs are failing under a certain conditions.
If you wish to see what's happening - let's do Bluejeans.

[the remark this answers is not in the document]
AlexanderZagaynov: How are you going to handle this trait? https://github.com/ansible/ansible_tower_client_ruby/blob/cf77e3bfa94222d3ddb1c59cd6eda332ec31f92b/spec/factories/responses.rb#L47
It makes that attribute a symbol.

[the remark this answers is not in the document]
bzwei: I now understand the cause of the above observations. In reality the constructor is not a hash but the JSON string from API response. The keys are strings not symbols. There are not mix uses of strings and symbols.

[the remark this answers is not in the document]
AlexanderZagaynov: And your test shows the same.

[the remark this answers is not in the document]
AlexanderZagaynov: And the main problem with specs:
When you run my test alone, it will work:
`$ bundle exec rspec spec/job_template_spec.rb:31`
https://github.com/ansible/ansible_tower_client_ruby/blob/cf77e3bfa94222d3ddb1c59cd6eda332ec31f92b/spec/job_template_spec.rb#L31
But if you run tests all together, it will fail, because of that string vs symbol mess.

If you'll comment that line:
https://github.com/ansible/ansible_tower_client_ruby/blob/cf77e3bfa94222d3ddb1c59cd6eda332ec31f92b/spec/job_template_spec.rb#L11
Or simply "xit" this line:
https://github.com/ansible/ansible_tower_client_ruby/blob/cf77e3bfa94222d3ddb1c59cd6eda332ec31f92b/spec/support/shared_examples/job_template.rb#L2
Test will pass.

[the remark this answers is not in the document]
AlexanderZagaynov: In reply to your: https://github.com/ansible/ansible_tower_client_ruby/pull/119#discussion_r211289385
My change only transforms all the keys to strings, and also transforms attribute names to strings too (to handle `jt.some_property`). So your algo will keep work as you wanted.

spec/job_template_spec.rb
```diff
@@ -16,14 +16,25 @@
     let(:json) { {'extra_vars' => "{\"instance_ids\":[\"i-999c\"],\"state\":\"absent\",\"subnet_id\":\"subnet-887\"}"} }
     let(:post_result_body) { {:job => 1} }
     let(:json_with_limit) { json.merge('limit' => 'machine_name') }
+    let(:response_double) { instance_double("Faraday::Response", :body => post_result_body.to_json) }
 
     it "runs an existing job template with a limit" do
-      launch_url      = "#{raw_instance["url"]}launch/"
-      response_double = instance_double("Faraday::Response", :body => post_result_body.to_json)
+      launch_url = "#{raw_instance["url"]}launch/"
+      raw_instance = build(:response_instance, :limitable_job_template, :klass => described_class)
```
bzwei: You can create two test variables in line 7
```
 let(:raw_instance_ask_limit)    { build(:response_instance, :job_template, :klass => described_class, :ask_limit_on_launch => true) }
 let(:raw_instance_no_limit)     { build(:response_instance, :job_template, :klass => described_class, :ask_limit_on_launch => false) }
```
Then in two test cases assert the following
```
expect { described_class.new(api, raw_instance_ask_limit.to_json).launch(json_with_limit) }.to_not raise_error
```
and
```
expect { described_class.new(api, raw_instance_no_limit.to_json).launch(json_with_limit) }.to raise_error ArgumentError
```

Here we want to convert the hash to JSON to mimic the response from API call. It also removes the confusion of strings vs symbols. No conversion is needed in the base model.

If you don't like to create `let` blocks, you can choose to add trait in the factory. The result will be the same.

lib/ansible_tower_client/base_models/job_template.rb
```diff
@@ -25,5 +37,21 @@ def extra_vars_hash
...
+    private
+
+    def check_ignorable_options(options)
+      ignored_options = IGNORABLE_OPTIONS.select do |option, checkmark|
+        !try(checkmark) && options.slice(option.to_sym, option.to_s).any?(&:present?)
+      end.keys
+
+      return if ignored_options.empty?
+
+      ignored_options.map! { |option| ":#{option}" }
```
Fryguy: I think you are just trying to get a printable version of the symbol here, in which case you can just use `.inspect`...

```ruby
ignored_options.map!(&:inspect)
```

Fryguy:
I had a similar thought to https://github.com/ansible/ansible_tower_client_ruby/pull/119#discussion_r210573230, which I think is still not resolved.  The method name implies "ignoring" the options, but then an ArgumentError is raised.  Instead of the "ignore" term, perhaps "validate" or "verify" ...for the constant maybe `INVALID_OPTIONS`?

spec/job_template_spec.rb
AlexanderZagaynov: Traits won't work, symbols won't work, as I told you here: https://github.com/ansible/ansible_tower_client_ruby/pull/119#discussion_r211287343
And you also mentioned strings vs symbols here: https://gitter.im/ManageIQ/manageiq?at=5b7adf72796f7b601d6b4dea

However, I solved that in an other way.

lib/ansible_tower_client/base_models/job_template.rb
AlexanderZagaynov: done

AlexanderZagaynov:
@Fryguy @gmcculloug @bdunne @bzwei PR was updated 3 days ago. Please review.

AlexanderZagaynov:
@Fryguy: Bronagh @bronaghs explained that **IGNORABLE_OPTIONS** sounds ambiguous.
We debate that, agreed on the semantical name, and I changed it to  **REQUIRED_ATTRIBUTES_FOR_OPTIONS**.

bronaghs:
@miq-bot assign @bdunne

lib/ansible_tower_client/base_models/job_template.rb
```diff
@@ -25,5 +37,20 @@ def extra_vars_hash
     def override_raw_attributes
       { :credential => :credential_id, :inventory => :inventory_id, :project => :project_id }
     end
+
+    private
+
+    def check_required_attributes_for_options(options)
```
bdunne: Maybe something like this would be better?
```ruby
def validate_launch_options(options)
  errors = []
  REQUIRED_ATTRIBUTES_FOR_OPTIONS.each do |option, checkmark|
    next unless (options.key?(option.to_sym) || options.key?(option.to_s)
    next if (respond_to?(checkmark) && send(checkmark))
    errors << "#{option} included but #{checkmark} was false"
  end

  return if errors.empty?

  raise ArgumentError, "Unused option(s): #{errors.join(", ")}"
end

```

bdunne: It also removes the need to Object#try, Object#blank and Hash#slice

lib/ansible_tower_client.rb
```diff
@@ -37,6 +37,11 @@
 
 require "more_core_extensions/all"
 require "active_support/inflector"
+require 'active_support/core_ext/object/try'
+require 'active_support/core_ext/object/blank'
+require 'active_support/core_ext/array/conversions'
+require 'active_support/core_ext/hash/slice'
+require 'active_support/core_ext/hash/except'
```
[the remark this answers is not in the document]
AlexanderZagaynov: Yes, in specs. However, that was already required via something else, so I'll just remove this line.

lib/ansible_tower_client.rb
```diff
@@ -37,6 +37,11 @@
 
 require "more_core_extensions/all"
 require "active_support/inflector"
+require 'active_support/core_ext/object/try'
+require 'active_support/core_ext/object/blank'
+require 'active_support/core_ext/array/conversions'
```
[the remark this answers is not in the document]
AlexanderZagaynov: It was used for https://github.com/ansible/ansible_tower_client_ruby/blob/2450d79a10e66d84f0c663b5619a6ce87af15a93/lib/ansible_tower_client/base_models/job_template.rb#L50
But now I removed that to satisfy your orders.

lib/ansible_tower_client/base_models/job_template.rb
AlexanderZagaynov: Some AS methods was already required via something else, so I'll just remove `try` and `to_sentence`.

AlexanderZagaynov:
@bdunne PR updated

spec/job_template_spec.rb
```diff
@@ -16,14 +16,25 @@
...
-      launch_url      = "#{raw_instance["url"]}launch/"
-      response_double = instance_double("Faraday::Response", :body => post_result_body.to_json)
+      launch_url = "#{raw_instance["url"]}launch/"
+      instance_data = raw_instance.merge('ask_limit_on_launch' => true)
+
       expect_any_instance_of(AnsibleTowerClient::Collection).to receive(:find).with(post_result_body[:job])
       expect(api).to receive(:post).with(launch_url, json_with_limit).and_return(response_double)
 
-      described_class.new(api, raw_instance).launch(json_with_limit)
+      expect { described_class.new(api, instance_data).launch(json_with_limit) }.to_not raise_error
```
bzwei: No need to specifically assert `to_not raise_error`. The test will fail if the method really raises an error.

lib/ansible_tower_client/base_models/job_template.rb
```diff
@@ -25,5 +37,20 @@ def extra_vars_hash
...
+    def validate_launch_options(options)
+      ignored_options = REQUIRED_ATTRIBUTES_FOR_OPTIONS.select do |option, checkmark|
+        options.slice(option.to_sym, option.to_s).any?(&:present?) && !(respond_to?(checkmark) && send(checkmark))
+      end.keys
+
+      return if ignored_options.empty?
+
+      message = ignored_options.map(&:inspect).join(', ')
+      message = ignored_options.size == 1 ? " #{message} is" : "s #{message} are"
+      message = "Option#{message} provided but corresponding ask on launch flag has not been turn on"
```
bzwei: The message needs some rephrase. I initially gave you some idea but not meant to be final words.

bdunne: Looks like my other comment got collapsed https://github.com/ansible/ansible_tower_client_ruby/pull/119#discussion_r213082218

AlexanderZagaynov: No I saw that. And that your proposal looks very ugly. However, if you insist...
But it will not be my code, it will be your code.

spec/job_template_spec.rb
AlexanderZagaynov: I know. But I prefer to use meaningful and semantically correct statements. If we test something - we should test (e.g. `expect` something), not just run and see what is going on.

lib/ansible_tower_client/base_models/job_template.rb
AlexanderZagaynov: @bzwei Please tell me what exactly you want here. I'm not a native English speaking person. I'm not a science fiction writer, or anything else like this.
Please don't ask me to "go somewhere I don't know where, and find something I don't know what".

spec/job_template_spec.rb
bzwei: Not a big deal here. You can leave it this way. We do have `expect` in line 25 and 26.

lib/ansible_tower_client/base_models/job_template.rb
gmcculloug: @AlexanderZagaynov Breaking up the message like this makes it difficult if we ever need to translate the string.

I prefer that we use a simple message that lists the flags as the end:
```ruby
raise ArgumentError, "'PROMPT ON LAUNCH' is required for the following fields: #{ignored_options.join(', ')}"
```
I removed the `map(&:inspect)` logic as well since I do not feel it is needed.


Note I am using the wording from the Tower UI for the checkbox name:
image

AlexanderZagaynov: Hi @gmcculloug! Thank you for such valuable comment!
It's fixed now.

AlexanderZagaynov:
@bdunne PR updated

bdunne:
@AlexanderZagaynov I released v0.18.0 including this change

AlexanderZagaynov:
Thank you @bdunne !

PR #122, opened 2018-11-09
Tower 3.3 removed result_stdout field from project_update

jameswnl:
Fixing https://github.com/ManageIQ/manageiq-automation_engine/issues/261

Related https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/140

### DIFF
--- a/lib/ansible_tower_client/base_models/project_update.rb
+++ b/lib/ansible_tower_client/base_models/project_update.rb
@@ -1,4 +1,8 @@
 module AnsibleTowerClient
   class ProjectUpdate < BaseModel
+    def stdout(format = 'txt')
+      api.get("#{related['stdout']}?format=#{format}").body
+    end
+    alias result_stdout stdout
   end
 end
--- a/spec/project_update_spec.rb
+++ b/spec/project_update_spec.rb
@@ -15,4 +15,20 @@ describe AnsibleTowerClient::ProjectUpdate do
     expect(obj.id).to   be_a Integer
     expect(obj.name).to be_a String
   end
+
+  context '#stdout' do
+    describe "exists" do
+      let(:stdout) { "Ansible Tower Project Update output" }
+
+      it "returns stdout default to plain text" do
+        expect(api).to receive(:get).with(/format=txt/).and_return(instance_double("Faraday::Result", :body => stdout))
+        expect(described_class.new(api, raw_instance).stdout).to eq(stdout)
+      end
+
+      it "returns formatted text per request" do
+        expect(api).to receive(:get).with(/format=html/).and_return(instance_double("Faraday::Result", :body => stdout))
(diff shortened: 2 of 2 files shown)


lib/ansible_tower_client/base_models/project_update.rb
```diff
@@ -1,4 +1,7 @@
 module AnsibleTowerClient
   class ProjectUpdate < BaseModel
+    def result_stdout
+      api.get(related['stdout']+'?format=txt').body
```
gmcculloug: I would prefer we default the format, but allow the caller to provider an override if needed.  Would be consistent with what we do for the Job class: https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/base_models/job.rb#L14

jameswnl:
@miq-bot remove_label wip
@miq-bot add_labels bug, hammer/yes

jameswnl:
@miq-bot assign @gmcculloug

lib/ansible_tower_client/base_models/project_update.rb
```diff
@@ -1,4 +1,8 @@
 module AnsibleTowerClient
   class ProjectUpdate < BaseModel
+    def stdout(format = 'txt')
+      api.get("#{related['stdout']}?format=#{format}").body
+    end
+    alias result_stdout stdout
```
bdunne: Does this need to be a version dependent patch?  Will this break older versions?

bdunne: Why add the `alias`?

jameswnl: no. I checked and this is available since 3.1

jameswnl: the miq tower gem is depending on this alias

bdunne: Do we still support versions older than 3.1?

jameswnl: Not sure.
At least 3.0 reached End of Maintenance Support 2 (EOL) on 2/28/18 according to 
https://access.redhat.com/support/policy/updates/ansible-tower

We started supporting Embedded Tower since 3.1. I am not sure about the external integration. Do you remember what version of Tower is the first integrated by MIQ?

gmcculloug: Since this change is intended to fix master and hammer branches I think we should be ok here based on the Tower EOL for 3.0.

jameswnl:
@gmcculloug @bdunne anything else needed for this to be merged?

gmcculloug:
I'm good with it.  @bdunne any issue merging?

jameswnl:
@bdunne can you create a tag to contain this fix?

bdunne:
@jameswnl v0.19.0 has been released

PR #123, opened 2019-02-07
v2 API compatibility

AlexanderZagaynov:
This PR allows to use Ansible Tower's v2 API, e.g. `https://example.com/api/v2`
It simply mimics old API's object fields to make new v2 objects compatible with v1.
It doesn't cover all the changes, only part of them, but it's enough to make https://github.com/ManageIQ/manageiq-providers-ansible_tower work.


lib/ansible_tower_client/api.rb
```diff
@@ -35,6 +35,10 @@ def credentials
       Collection.new(self, credential_class)
     end
 
+    def credential_types
+      Collection.new(self, credential_type_class)
```
jameswnl: so in a `v1` case, what is the expected behavior when accessing this?

jameswnl:
@Fryguy @bdunne take a look at this.  This will allow existing MIQ and the https://github.com/ManageIQ/manageiq-providers-ansible_tower to continue to work with Tower's `v2` API

lib/ansible_tower_client/base_models/credential.rb
```diff
@@ -1,7 +1,43 @@
...
+      'azure_rm'   => 'Microsoft Azure Resource Manager',
+      'openstack'  => 'OpenStack',
+      'rhv'        => 'Red Hat Virtualization',
+      'insights'   => 'Insights',
+      'tower'      => 'Ansible Tower',
+    }.invert.freeze
+
+    # https://github.com/ansible/awx/blob/1328fb80a02ef4e37bc021eb07d4be041a41f937/awx/main/models/credential/__init__.py#L301
+    def kind
+      @data['kind'] ||= begin
```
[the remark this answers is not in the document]
AlexanderZagaynov: Yes, goal is to keep behavior of v1 objects.

lib/ansible_tower_client/base_models/credential.rb
```diff
@@ -1,7 +1,43 @@
...
+        kind = credential_type.kind
+        kind == 'cloud' && KIND_CHOICES[credential_type.name] || kind
```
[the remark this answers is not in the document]
AlexanderZagaynov: you answered to that below :)

lib/ansible_tower_client/api.rb
AlexanderZagaynov: Since there is no such endpoint in v1, user will get 404 on http operations.
This is normal, because if user wants to work with v1, he should not use this object.

lib/ansible_tower_client/base_models/credential.rb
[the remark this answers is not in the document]
jameswnl: I don't get it. can you explain?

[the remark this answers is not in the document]
AlexanderZagaynov: For credential kind "cloud" there is several choices, like "aws", "azure", etc.
We need to get exact type, not just generic "cloud".

jrafanie:
Interesting.  I'll have to review more but we should probably gets the bug fixes #126 and #127 merged and released and then we can see about getting this merged.   In terms of risk, I'd rather we get the less risky changes in first, then give us time to test this PR and master before needing to release.

> It doesn't cover all the changes, only part of them,

We might even want to consider an api/v2 branch depending on how much more there is to do.  Thoughts @AlexanderZagaynov @jameswnl

AlexanderZagaynov:
@jrafanie here is some work in progress about pure new version from a scratch: https://github.com/AlexanderZagaynov/ansible_tower_client_ruby_v2
I've forwarded you emails with discussion about that.

bronaghs:
hi @jrafanie 
https://github.com/ansible/ansible_tower_client_ruby/pull/126 and https://github.com/ansible/ansible_tower_client_ruby/pull/127 have been merged and, I am told, released. Can this PR be merged?

jrafanie:
Yes, those PRs were merged and released for ruby 2.5 compatibility.  I'm not familiar with the v2 changes to feel confident either way.  The only concern is if we need more changes on the last stable version and we need to backport a change, we might need to create a branch at a prior release to do that backport.

mfeifer:
@jrafanie who can help with confidence level, here, so that this can move forward.

jrafanie:
@mfeifer someone who's familiar with the ansible provider in both embedded and standalone form who knows how we represent the ansible objects via refresh and through operations... it's unclear if this works with all supported versions and what testing was done.

mfeifer:
@jrafanie and do you have any people in mind? I don't.

jrafanie:
Looking at the committers other than Brandon, maybe @syncrou or @jameswnl know more of the tower api side ... @agrare from a general providers perspective.  It would also be hopeful to know what versions we need to support in master for ansible tower and if this PR was tested for these versions.  Maybe @AlexanderZagaynov can update this PR with some information about that.  Also, it's unclear if this is really important because v1 is going away or if it's a general enhancement, as there is no bugzilla or linked task to do this work.  🤷‍♂️

AlexanderZagaynov:
Hi Joe, apologies, I didn't put a links in description.
Here is the BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1677553

We started that with @jameswnl, and he were curating my work there, he has all the knowledge, and I was expecting him to merge this PR.

Let me write some brief here:
- This change doesn't affects `v1` interactions, everything will work as before (both embedded and tower).
- As I know, Ansible Tower going to drop support for `v1` very soon.
- This change tries to get `v1` attributes, and if they missing - takes their values from `v2` objects if they available.
- User should point to `v2` endpoint specially if wants to use it. Otherwise it works as before: if URL has host only - it uses `v1` path (as far as I know, I didn't touch that, just saw somewhere in the code). If user enters URL with `v2` path - we use `v2`.

AlexanderZagaynov:
I tested everything manually, and @jameswnl is a witness here. Don't ask me to do it again please, it took a bunch of time :)

jameswnl:
hey, yeah, so like my previous comments, this is a pretty smart tweak to make the gem to work with v2. But my concern is that this may not be an obvious approach (or fragile).  So I would like to hear from @Fryguy and @bdunne

BTW, I don't have merge right for this repo.

jrafanie:
If this just works, I wouldn't be opposed to getting it in but I'm not sure what a longer term goals are here and if the changes here would be ok for now.

AlexanderZagaynov:
@jrafanie sorry, missed your reply. As I see it - long term goal is to rewrite this gem and refactor related MiQ code, I posted link above.

AlexanderZagaynov:
@mfeifer looks like things haven't budged an inch? :smiley:

mfeifer:
@Fryguy any thoughts on how to proceed with this one?

just in case you want to follow it: https://projects.engineering.redhat.com/browse/CFMEDEV-30

lib/ansible_tower_client/base_models/credential.rb
bdunne: Prefer ternary operator for readability

lib/ansible_tower_client/base_models/project.rb
```diff
@@ -19,8 +19,19 @@ def update
     end
 
     def last_update
-      return if related['last_update'].blank?
-      api.project_updates.find(related['last_update'])
+      return @last_update if defined? @last_update
+      return @last_update = nil unless related.raw_hash.key?('last_update')
```
bdunne: If the key is not there now, it won't be populated later?

lib/ansible_tower_client/base_models/project.rb
```diff
@@ -19,8 +19,19 @@ def update
...
+      return @last_update = nil if (update_id = related.last_update).blank?
```
bdunne: Same here?

lib/ansible_tower_client/base_models/project.rb
```diff
@@ -19,8 +19,19 @@ def update
...
+
+      unless update_id.kind_of?(Numeric) || update_id =~ /\A\d+\z/
+        if raw_hash.key?('summary_fields') && summary_fields.raw_hash.key?('last_update')
+          update_id = summary_fields.last_update.id
+        else
+          /\/(?'update_id'\d+)\/?\z/ =~ update_id
```
bdunne: Do you have an example of this string?  It might be nice to have as a comment or a test

lib/ansible_tower_client/base_models/project.rb
```diff
@@ -19,8 +19,19 @@ def update
...
+        end
+      end
+
+      @last_update = update_id && api.project_updates.find(update_id)
```
bdunne: If `update_id` is `nil` it will not be populated in the future either?

lib/ansible_tower_client/base_models/project.rb
AlexanderZagaynov: This is simple named group capture of all digits after the last `/`, optionally ending on the very last `/`.
E.g., this is url like `http://abc.def/something/12345/`

lib/ansible_tower_client/base_models/project.rb
AlexanderZagaynov: Yes, `defined?` everywhere exactly for this purpose.

lib/ansible_tower_client/base_models/project.rb
AlexanderZagaynov: No, if it was once nil, it will be nil further. That's why I check with `defined?`. This way I'm preventing multiple API requests.

lib/ansible_tower_client/base_models/credential.rb
AlexanderZagaynov: Ternary do not suit here well, so I had to add default value to `KIND_CHOICES` hash few lines above, now logic works the same as it was expected. Done.

lib/ansible_tower_client/base_models/project.rb
AlexanderZagaynov: Yes, `defined?` above does the trick.

lib/ansible_tower_client/base_models/project.rb
```diff
@@ -19,8 +19,19 @@ def update
     end
 
     def last_update
-      return if related['last_update'].blank?
-      api.project_updates.find(related['last_update'])
+      return @last_update if defined? @last_update
+      return @last_update = nil unless related.raw_hash.key?('last_update')
+      return @last_update = nil if (update_id = related.last_update).blank?
+
+      unless update_id.kind_of?(Numeric) || update_id =~ /\A\d+\z/
```
Fryguy: Minor, but I find the `unless ... || ...` construct confusing; this might be more readable as `if !update_id.kind_of?(Numeric) && update_id !~ /\A\d+\z/`

(Even that way, I find this conditional hard to read.)

AlexanderZagaynov: Two _nots_ in _if_, I guess rubocop will not be pleased? If it's hard for read anyway, I have an idea how to make you happier (I hope :smile:).

AlexanderZagaynov: done, commit https://github.com/ansible/ansible_tower_client_ruby/pull/123/commits/bd12dabe00f8eb4b4e0c4be680b4202778c9ffe0

AlexanderZagaynov: please tell me, who is boss here, you or rubocop? :smiley:  https://github.com/ansible/ansible_tower_client_ruby/pull/123#issuecomment-494736702

lib/ansible_tower_client/base_models/project.rb
bdunne: Yes, I can mentally parse the regex, but it is only a small part of a fairly complicated method, which really should have tests.  My point is that it's much easier to add tests for a feature / method when it is being written that describe the expected inputs and outputs than for the developer fixing a bug to have to discover (or guess) all of the cases that need to be covered.

bdunne:
Version 0.20.0 has been released https://rubygems.org/gems/ansible_tower_client/versions/0.20.0

simaishi:
Is this safe to go to `hammer` branch?

We need https://github.com/ansible/ansible_tower_client_ruby/pull/131 to be included in hammer and I was reviewing changes that went into this gem after 1.19.1 release (the version used in the last build). This PR talks about concerns around future backport, so I wanted to check.

cc @dmetzger57 @smallamp

AlexanderZagaynov:
@simaishi yes, those changes are backwards compatible

PR #124, opened 2019-02-09
little improvements

AlexanderZagaynov:
**avoid redundant redirects**

Currently, base resource paths got requested without trailing slashes in addresses.
It leads so such lines in the logs:
```
INFO -- : get https://example.com/api/v2/config
INFO -- Status: 301
INFO -- : get https://example.com/api/v2/config/
INFO -- Status: 200
INFO -- : get https://example.com/api/v2/inventories
INFO -- Status: 301
INFO -- : get https://example.com/api/v2/inventories/
INFO -- Status: 200
```
This commit eliminates useless redirect hops.

**unhide url of not found resource**

Ease `404`-related errors debugging.

### DIFF
--- a/spec/support/mock_api.rb
+++ b/spec/support/mock_api.rb
@@ -1,11 +1,32 @@
 module AnsibleTowerClient
   class MockApi
     class Response
-      attr_reader :body
-      def initialize(body)
+      attr_reader :body, :url
+
+      def initialize(body, url = nil)
         @body = body
+        @url  = url
       end
     end
+
+    RESPONSES = {
+      'ad_hoc_commands'             => 'AdHocCommand',
+      'config'                      => 'Config',
+      'credentials'                 => 'Credential',
+      'groups'                      => 'Group',
+      'hosts'                       => 'Host',
+      'inventories'                 => 'Inventory',
+      'inventory_sources'           => 'InventorySource',
+      'jobs'                        => 'Job',
+      'job_templates'               => 'JobTemplate',
+      'organizations'               => 'Organization',
+      'projects'                    => 'Project',
+      'me'                          => 'Me',
+      'workflow_jobs'               => 'WorkflowJob',
+      'workflow_job_nodes'          => 'WorkflowJobNode',
+      'workflow_job_template_nodes' => 'WorkflowJobTemplateNode',
+      'workflow_job_templates'      => 'WorkflowJobTemplate',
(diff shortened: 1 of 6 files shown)


AlexanderZagaynov:
@bdunne @jameswnl please review

spec/support/mock_api.rb
```diff
@@ -1,56 +1,48 @@
...
+      attr_reader :body, :url
+
+      def initialize(body, url = nil)
         @body = body
+        @url  = url
       end
     end
 
+    RESPONSES = {
+      'ad_hoc_commands'             => 'AdHocCommand',
```
jameswnl: too bad (or good) that we are not in rails, otherwise, this hash can be replaced by `classify`

AlexanderZagaynov: considering that previously it was `case` chain instead of simple hash table / associative array, I believe it's a huge improvement anyway :)

jameswnl:
@bdunne can you take a look?

jrafanie:
@AlexanderZagaynov this change will force VCR cassettes to be manually sed/changed or we'll have to re-record them for the ansible provider.  This is because we previously recorded gets for `config` which will now be `config/` and likewise for all collections.

AlexanderZagaynov:
@jrafanie makes sense. I also noticed one more place with missing trailing slash: https://github.com/ansible/ansible_tower_client_ruby/blob/5ec38ba8ebab432f22d5daeb86ca09c2bcee3777/lib/ansible_tower_client/api.rb#L23

jrafanie:
@AlexanderZagaynov did you want to get that change in before we merge/release so we can re-record or sed/awk the vcr cassettes only once?

jrafanie:
@AlexanderZagaynov Ruby 2.5 support doesn't need to happen today or tomorrow.. it can wait until we have all of our changes we want into the gem.  If so, please open a PR so we can try releasing a gem version next week.  Thanks!

AlexanderZagaynov:
> did you want to get that change in before we merge/release so we can re-record or sed/awk the vcr cassettes only once?
if you mean that `me/` redirect, I've prepared a new PR for this, mentioned above.

> Ruby 2.5 support doesn't need to happen today or tomorrow.. it can wait until we have all of our changes we want into the gem. If so, please open a PR so we can try releasing a gem version next week. Thanks!
done :)

PR #126, opened 2019-03-19
[ruby 2.5] Ensure no double trailing slash

jrafanie:
Related project updates perform a find with an id with a trailing slash,
such as 'project_updates/233/'.  Previously, we'd append another
trailing slash and it would be up to the underlying faraday code to fix it.

On ruby 2.5.x, faraday would receive two trailing slashes and not fix
it.  On ruby 2.4.x, we'd send two trailing slashes to faraday
but it would perform the get with only a single trailing slash.

It's unclear what changed in ruby 2.4 vs. 2.5 to behave differently with the same
faraday and other gems, but it's better to ensure we don't try to send an invalid
URI to faraday.

### DIFF
--- a/lib/ansible_tower_client/collection.rb
+++ b/lib/ansible_tower_client/collection.rb
@@ -28,7 +28,7 @@ module AnsibleTowerClient
     end
 
     def find(id)
-      build_object(parse_response(api.get("#{klass.endpoint}/#{id}/")))
+      build_object(parse_response(api.get(find_uri(id))))
     end
 
     def create!(*args)
@@ -49,6 +49,10 @@ module AnsibleTowerClient
       return if next_page.nil?
       body = parse_response(api.get(next_page, get_options))
       parse_result_set(body)
+    end
+
+    def find_uri(id)
+      File.join(klass.endpoint, id.to_s, "/")
     end
 
     def parse_response(response)
--- a/spec/collection_spec.rb
+++ b/spec/collection_spec.rb
@@ -55,6 +55,14 @@ describe AnsibleTowerClient::Collection do
         expect(mock_api.send(collection_class).find(1)).to be_instance_of(klass)
       end
     end
+
+    it "#find_uri(private)" do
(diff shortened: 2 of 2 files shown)


jrafanie:
@AlexanderZagaynov are you familiar with recording cassettes for the ansible provider?

If so, your changes mentioned [here](https://github.com/ansible/ansible_tower_client_ruby/pull/124#issuecomment-474602705) will need new recordings when we want to use the next version that includes my changes here.

juliancheal:
LGTM 👍

AlexanderZagaynov:
@jrafanie you mean, we need PR with all recordings refreshed? 
If so, I can do that after current one will be merged.

jrafanie:
@AlexanderZagaynov yes, we'll need to re-record the vcr cassettes or do a clever awk/sed script to update them to include the trailing `/` for all of the requests.  once we have all the changes in, we can release the gem and upgrade when we have the cassettes ready.

jrafanie:
@Fryguy please review 🙏

lib/ansible_tower_client/collection.rb
```diff
@@ -28,7 +28,11 @@ def find_all_by_url(url, get_options = nil)
     end
 
     def find(id)
-      build_object(parse_response(api.get("#{klass.endpoint}/#{id}/")))
+      build_object(parse_response(api.get(find_uri(id))))
+    end
+
+    def find_uri(id)
+      File.join(klass.endpoint, id.to_s, "/")
```
Fryguy: Should this method be private

spec/collection_spec.rb
```diff
@@ -55,6 +55,14 @@
         expect(mock_api.send(collection_class).find(1)).to be_instance_of(klass)
       end
     end
+
+    it "find_uri" do
```
Fryguy: Should have a leading #

Fryguy:
A couple minor things

jrafanie:
Ok, fixed up the suggestions and cleaned up the commit message/description.  Thanks @Fryguy

PR #127, opened 2019-03-22
avoid redundant redirects

AlexanderZagaynov:
this is a follow-up for https://github.com/ansible/ansible_tower_client_ruby/pull/124
found one more place with redirecting url


AlexanderZagaynov:
@bdunne @jrafanie please take a look

PR #128, opened 2019-04-19
Update bundler, rake, remove unnecessary load path mangling

djberg96:
This updates the gemspec by updating the bundler and rake versions, removes some unnecessary load path mangling, and gets rid of the `require_paths` specification since 'lib' is already the default.

Edit: also realized factory_girl needed to be updated to factory_bot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: ruby
 rvm:
-  - "2.3.1"
-  - "2.4.1"
+  - "2.4.5"
+  - "2.5.3"
 sudo: false
 cache: bundler
-before_install: gem install bundler -v 2.0.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,4 @@ rvm:
   - "2.4.1"
 sudo: false
 cache: bundler
-before_install: gem install bundler -v 1.10.3
+before_install: gem install bundler -v 2.0.1
--- a/ansible_tower_client.gemspec
+++ b/ansible_tower_client.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "faraday_middleware"
   spec.add_runtime_dependency "more_core_extensions", "~> 3.0"
 
-  spec.add_development_dependency "factory_bot", "~> 4.11.1"
-  spec.add_development_dependency "rake", "~> 12.3.2"
+  spec.add_development_dependency "factory_bot", "~> 4.11"
+  spec.add_development_dependency "rake", "~> 12.3"
   spec.add_development_dependency "rspec"
 end
--- a/ansible_tower_client.gemspec
+++ b/ansible_tower_client.gemspec
@@ -21,7 +21,6 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "faraday_middleware"
   spec.add_runtime_dependency "more_core_extensions", "~> 3.0"
 
-  spec.add_development_dependency "bundler", "~> 2.0.1"
   spec.add_development_dependency "factory_bot", "~> 4.11.1"
   spec.add_development_dependency "rake", "~> 12.3.2"
   spec.add_development_dependency "rspec"
--- a/ansible_tower_client.gemspec
+++ b/ansible_tower_client.gemspec
@@ -1,7 +1,5 @@
 # coding: utf-8
-lib = File.expand_path('../lib', __FILE__)
-$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
-require 'ansible_tower_client/version'
+require_relative 'lib/ansible_tower_client/version'
 
 Gem::Specification.new do |spec|
   spec.name          = "ansible_tower_client"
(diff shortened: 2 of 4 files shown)


djberg96:
@Fryguy @bdunne Please review.

djberg96:
Note that I had to update the dynamic attribute syntax.

Also note that with factory_bot > 5 there's a failure on `instance_id` being an undefined method, so I've pinned it to 4.11 for now.

djberg96:
@miq-bot add_reviewer @syncrou

ansible_tower_client.gemspec
```diff
@@ -1,7 +1,5 @@
 # coding: utf-8
-lib = File.expand_path('../lib', __FILE__)
-$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
-require 'ansible_tower_client/version'
+require_relative 'lib/ansible_tower_client/version'
```
syncrou: Is `require_relative` always safe? Is there a reason we're not following the
```
$:.push File.expand_path("../lib", __FILE__)
```
Method used in our other gems?

djberg96: Not sure what you mean by "safe" here. It has the advantage of not mangling `$LOAD_PATH`, plus it looks nicer.

djberg96: The `require_relative` method was added at some point in the Ruby 2.x lifecycle, and I think a lot of code predates that, or simply copy-pasta's the older code.

gmcculloug: @jrafanie Do you know if we prefer one style or `require` over another here?

jrafanie: I think `require_relative` is fine.  I don't see why we can't use it here.  

I think the default behavior of adding a gem to your Gemfile is it tries to put the gem's lib directory on the load path.  So, I think the load path change here was redundant.  So, I'm 👍

jrafanie: I don't know why we did the load path stuff here though.  😨

.travis.yml
```diff
@@ -4,4 +4,4 @@ rvm:
   - "2.4.1"
 sudo: false
 cache: bundler
-before_install: gem install bundler -v 1.10.3
+before_install: gem install bundler -v 2.0.1
```
jrafanie: is this necessary anymore?  Does it work without it?  I think we only had it because we didn't want a new bundler to be installed... but we are specifying a new bundler so maybe it's no longer needed?

ansible_tower_client.gemspec
```diff
@@ -17,15 +15,14 @@ Gem::Specification.new do |spec|
   spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
   spec.bindir        = "exe"
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
-  spec.require_paths = ["lib"]
```
jrafanie: hmm, why remove this?

djberg96: It's already the default: https://guides.rubygems.org/specification-reference/#require_paths=

jrafanie: 🤷‍♂️ why do we do these things?  Lol

jrafanie: maybe it wasn't the default at one time or we had more than just lib...

djberg96: I see this line copied around a lot in gemspecs out of habit I think.

.travis.yml
djberg96: Oh, you mean just remove the line completely? I can try it. Wondered why it was there.

djberg96: Ok, took that out, though it required removing it as a dev dependency, which should be fine (you usually don't specify that anyway). I also bumped the rvm versions to match manageiq.

djberg96: @jrafanie ok, green again. :)

ansible_tower_client.gemspec
Fryguy: > why do we do these things? Lol

We did this because gem builders do this...same with the other changes in this PR

Fryguy:
Just to extend on the comment thread here: https://github.com/ansible/ansible_tower_client_ruby/pull/128#discussion_r278329261

We didn't do this, bundler did as it was used for gem creation, and most of the changes here undo the bundler default way (not saying we shouldn't merge, just clarifying).  Try it locally with `bundle gem foo_bar` and you can see it still creates the `$LOAD_PATH` and `require_paths` lines from the gemspec, and the `before_install` line in .travis.yml.

djberg96:
I'm actually going to make one minor modification, which is to make the deps less pessimistic.

gmcculloug:
@jrafanie Are you good with these last changes?  Comment https://github.com/ansible/ansible_tower_client_ruby/pull/128/commits/2c270f01fc744ef982f9f8509aa7cff7e2e4aa9f

jrafanie:
@gmcculloug yeah, those are dev dependencies.  I'm good with making them less strict.

PR #134, opened 2019-11-27
Support for headers and proxy

slemrmartin:
Added support for proxy requests and http headers


slemrmartin:
@bdunne please review

PR #139, opened 2020-02-04
[WIP] Add upper bounds to faraday versions

d-m-u:
It feels like we maybe might want something like this? 🤷‍♀ 

But we're waiting on them to cut a new middleware gem:
v0.14 is `'faraday', ['>= 0.7.4', '< 1.0']` and won't work

### DIFF
--- a/ansible_tower_client.gemspec
+++ b/ansible_tower_client.gemspec
@@ -16,9 +16,9 @@ Gem::Specification.new do |spec|
   spec.bindir        = "exe"
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
 
-  spec.add_runtime_dependency "activesupport"
-  spec.add_runtime_dependency "faraday"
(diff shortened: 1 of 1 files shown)


ansible_tower_client.gemspec
```diff
@@ -17,8 +17,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
 
   spec.add_runtime_dependency "activesupport"
-  spec.add_runtime_dependency "faraday"
```
bdunne: It looks like v1.0.0 was released a couple days ago, do we need to do anything else to upgrade to it?  Also, if 1.0 is okay, we should probably use `~>1.0`

d-m-u:
Closing because I expect that with the license change to vcr we can't use Faraday 1.0.0 anyway. Per conversations in gitter recently and with Jason the new license will prohibit RH from using the latest VCR and so this update isn't going to happen.

PR #144, opened 2020-06-09
Return connection error if not able to parse input

kbrock:
Detect invalid json response from the server

### DIFF
--- a/lib/ansible_tower_client/api.rb
+++ b/lib/ansible_tower_client/api.rb
@@ -20,7 +20,10 @@ module AnsibleTowerClient
     end
 
     def verify_credentials
-      JSON.parse(get("me/").body).fetch_path("results", 0, "username")
+      results = get("me/").body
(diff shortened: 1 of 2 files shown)


lib/ansible_tower_client/api.rb
```diff
@@ -20,7 +20,10 @@ def version
     end
 
     def verify_credentials
-      JSON.parse(get("me/").body).fetch_path("results", 0, "username")
+      results = get("me/").body
+      JSON.parse(results).fetch_path("results", 0, "username")
+    rescue JSON::ParserError
+      raise AnsibleTowerClient::ConnectionError, "invalid host"
```
bdunne, line 26: Why `"invalid host"`?  That makes me think 4XX error.  Would "unexpected response" make more sense?

kbrock: thanks @bdunne

PR #147, opened 2021-06-02
[WIP] Improve handling of nested workflow_jobs

agrare:
Improve the handling of nested workflow_jobs in the WorkflowJobNode class.

The `#job` method was using `.match?(/jobs/)` to check if it was a job, the problem is that was matching on `/api/v2/workflow_jobs/:id` as well causing `api.jobs.find(job_id)` to fail with a 404.

Fix for the issue noted here: https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/262#issuecomment-853317690

TODO:
- [x] Add spec tests for `#workflow_job`

### DIFF
--- a/lib/ansible_tower_client/base_models/workflow_job_node.rb
+++ b/lib/ansible_tower_client/base_models/workflow_job_node.rb
@@ -5,14 +5,20 @@ module AnsibleTowerClient
     end
 
     def job
-      api.jobs.find(job_id) if job?
+      return unless job?
+
+      if related.job.include?("/jobs/")
+        api.jobs.find(job_id)
+      elsif related.job.include?("/workflow_jobs/")
+        api.workflow_jobs.find(job_id)
+      end
     end
 
     # to filter out WorkflowJobNode that is inventory sync or project sync
     def job?
       return false if !respond_to?(:job_id) || job_id.nil?
 
-      related.job.match?(/jobs/)
+      related.job.match?(/\/(jobs|workflow_jobs)\//)
     end
   end
 end
--- a/spec/workflow_job_node_spec.rb
+++ b/spec/workflow_job_node_spec.rb
@@ -40,6 +40,13 @@ describe AnsibleTowerClient::WorkflowJobNode do
       expect(obj.job).to be_a AnsibleTowerClient::Job
     end
 
+    it "returns a WorkflowJob when the job_id has a workflow_job" do
+      obj = described_class.new(api, response_with_job(raw_instance, "/api/v1/workflow_jobs/2710/"))
+      allow(api).to receive(:get).and_return(instance_double("Faraday::Result", :body => {}.to_json))
+
+      expect(obj.job).to be_a AnsibleTowerClient::WorkflowJob
+    end
+
     it "returns nil when it is a inventory sync" do
       obj = described_class.new(api, response_with_job(raw_instance, "/api/v1/inventory_updates/2710/"))
       allow(api).to receive(:get).and_return(instance_double("Faraday::Result", :body => {}.to_json))


lib/ansible_tower_client/base_models/workflow_job_node.rb
```diff
@@ -8,11 +8,21 @@ def job
       api.jobs.find(job_id) if job?
     end
 
+    def workflow_job
```
chessbyte, line 11: what calls this method?

agrare: This is a public interface, the equivalent of `#job` up on line 7 but for `workflow_job` types

agrare:
For reference this is what one of these workflow_job_nodes looks like with a nested workflow_job
```
{
    "id": 11,
    "type": "workflow_job_node",
    "url": "/api/v2/workflow_job_nodes/11/",
    "related": {
        "credentials": "/api/v2/workflow_job_nodes/11/credentials/",
        "success_nodes": "/api/v2/workflow_job_nodes/11/success_nodes/",
        "failure_nodes": "/api/v2/workflow_job_nodes/11/failure_nodes/",
        "always_nodes": "/api/v2/workflow_job_nodes/11/always_nodes/",
        "unified_job_template": "/api/v2/workflow_job_templates/9/",
        "job": "/api/v2/workflow_jobs/27/",
        "workflow_job": "/api/v2/workflow_jobs/26/"
    },
    "summary_fields": {
        "job": {
            "id": 27,
            "name": "Test Workflow Template",
            "description": "",
            "status": "successful",
            "failed": false,
            "elapsed": 7.663,
            "type": "workflow_job"
        },
        "workflow_job": {
            "id": 26,
            "name": "Test Nested Workflow",
            "description": ""
        },
        "unified_job_template": {
            "id": 9,
            "name": "Test Workflow Template",
            "description": "",
            "unified_job_type": "workflow_job"
        }
    },
    "created": "2021-05-24T19:55:57.183220Z",
    "modified": "2021-05-24T19:55:57.439811Z",
    "extra_data": {},
    "inventory": null,
    "scm_branch": null,
    "job_type": null,
    "job_tags": null,
    "skip_tags": null,
    "limit": null,
    "diff_mode": null,
    "verbosity": null,
    "job": 27,
    "workflow_job": 26,
    "unified_job_template": 9,
    "success_nodes": [],
    "failure_nodes": [],
    "always_nodes": [],
    "all_parents_must_converge": false,
    "do_not_run": false,
    "identifier": "0e4519d5-a7f1-41b0-ad17-a2a8a140fd91"
}
```

agrare:
cc @bdunne

lib/ansible_tower_client/base_models/workflow_job_node.rb
```diff
@@ -12,7 +12,13 @@ def job
     def job?
       return false if !respond_to?(:job_id) || job_id.nil?
 
-      related.job.match?(/jobs/)
+      related.job.include?("/jobs/")
```
agrare, line 15: This fixes the issue where a nested workflow job would return that `job?` was true, because `"/api/v2/workflow_jobs/1234".match?(/jobs/)` is `true`, to fix this it seems safer to look for `"/jobs/"` instead of just `"jobs"`

lib/ansible_tower_client/base_models/workflow_job_node.rb
```diff
@@ -5,14 +5,20 @@ def workflow_job
...
 
     def job
-      api.jobs.find(job_id) if job?
+      return unless job?
+
+      if related.job.include?("/jobs/")
+        api.jobs.find(job_id)
+      elsif related.job.include?("/workflow_jobs/")
+        api.workflow_jobs.find(job_id)
+      end
```
agrare, line 14: Unsure how best to handle this, should `job` return `Job` and `WorkflowJob` types?  That would match the behavior of the tower API where there is a workflow_job id in the `job_id` property.  @bdunne opinions?

